PR #52, opened 2020-08-10
Fix field.sub_fields to return results for nested table type

vdboor:
This logic was written in the 'dso docs', and now upstreamed.

### DIFF
--- a/schematools/types.py
+++ b/schematools/types.py
@@ -57,7 +57,18 @@ class DatasetSchema(SchemaType):
         return cls(obj)
 
     @property
+    def title(self):
+        """Title of the dataset (if set)"""
+        return self.get("title")
+
+    @property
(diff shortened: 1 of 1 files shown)


schematools/types.py
```diff
@@ -274,26 +289,54 @@ def nm_relation(self) -> typing.Optional[str]:
...
+                    _required=(name in required),
+                    **spec,
+                )
+                for name, spec in self["properties"].items()
+            ]
+        elif self.is_array:
+            # Also consider the sub-fields for a M2M relationship.
+            # These can also be used for filtering (used in docs).
+            items = self.items
+            if items and items.get("type") == "object":
```
ngaranko, line 326: Array van objects is een "nested table" en geen M2M relatie.

PR #61, opened 2020-10-05
Fix the fields for an nm-relation

jjmurre:
Make names more specific for both sides of the relation.
This avoids name-clashes for self-referencing nm-relations

Add the left-side values of an nm-relation for compound keys

### DIFF
--- a/schematools/types.py
+++ b/schematools/types.py
@@ -235,6 +235,11 @@ class DatasetTableSchema(SchemaType):
                 _name="id", _parent_table=self, _required=True, type="string"
             )
 
+    def get_fields_by_id(self, field_names) -> typing.List[DatasetFieldSchema]:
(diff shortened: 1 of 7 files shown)


schematools/types.py
```diff
@@ -235,6 +235,11 @@ def fields(self):
                 _name="id", _parent_table=self, _required=True, type="string"
             )
 
+    def get_fields_by_id(self, field_names) -> typing.List[DatasetFieldSchema]:
```
vdboor, line 238:
```suggestion
    def get_fields_by_id(self, field_names) -> typing.Generator[DatasetFieldSchema]:
```

jjmurre: Goeie, tx!

bheupers (approved):
Ziet er goed uit met test :)

PR #64, opened 2020-10-20
Add a custom field type to use for "loose" relations

jjmurre:
Those loose relations just store a key value to be used in
the construction of a relation url in the DSO, without any
checking on integrity.

This field type can be used to have relations to temporal datasets
(using identificatie/volgnummer as key) purely based on identificatie alone.


bheupers (approved):
Lijkt me goed

PR #69, opened 2020-10-24
Changes to schema-tools for BRP mandatory-queries field

vdboor:
* Added `ProfileSchema` object, which is also exposed through `Profile` model.
* Added properties to make the profile schema easier to access in code (like the `DatasetTableSchema` and friends do).
* Fixed some bugs


vdboor:
Part of #75

PR #87, opened 2020-12-21
[DS-514] DB_NAME instance method for getting table name

cvriel:
In types.py a DB_NAME instance method is used for getting the DB name of a table.
In the future, it can be used to look up the SHORTNAME in the Amsterdam Schema definition
to accomodate table names that can exceed the Postgres DB limit of 63 characters.

The DB_NAME instance method must also deal with n:m tables (through tables).
The Index name creation on n:m tables uses the DB_NAME method to look up the table name to apply its index.

Defining the DB_NAME instance method in types.py on - at first - class DatasetTableSchema, set a placeholder for a the Amsterdam Schema look up in the near future.
For now, it uses an existing function that defines the table name which is also used in the baggob data processing.

### DIFF
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -1,10 +1,6 @@
 from schematools.types import SchemaType, DatasetSchema
 from schematools.importer.base import BaseImporter
-from schematools.cli import _get_engine
-import environ
 from sqlalchemy import MetaData, create_engine, inspect
-
-env = environ.Env()
 
 
 def test_index_creation(engine, db_schema):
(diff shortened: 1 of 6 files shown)


tests/test_indexes.py
```diff
@@ -154,29 +152,28 @@ def test_index_troughtables_creation():
...
+        )
 
     for table in data["tables"]:
 
         dataset_table = dataset_schema.get_table_by_id(table["id"])
 
         for table in dataset_table.get_through_tables_by_id():
 
-            CONN = create_engine(DATABASE_URL, client_encoding="UTF-8")
+            CONN = create_engine(engine.url, client_encoding="UTF-8")
```
jjmurre, line 173: Namen van variabele in all caps betekent in python dat het constanten zijn, maar dat is hier niet het geval, ik zou er gewoon lowercase (snakecase) van maken.

jjmurre (approved):
Lijkt me mooi zo!

PR #88, opened 2020-12-22
Make temporal dimension fields optional

jjmurre:
Temporal dimension fields (begin_geldigheid, eind_geldigheid) were inserted to the Django models (for temporary dataset-tables) by default. However, most of the current database tables do not have these fields, because the current ndjson import does not provide these fields. The generation of dimension fields is now made configurable and switched off by default.


cvriel (approved):
Prima om dat optioneel te triggeren => begin en eindgeldigheid

PR #90, opened 2020-12-23
[DS-514] DB_NAME instance method for getting table name

cvriel:
In types.py a DB_NAME instance method is used for getting the DB name of a table.
In the future, it can be used to look up the SHORTNAME in the Amsterdam Schema definition
to accomodate table names that can exceed the Postgres DB limit of 63 characters.
    
The DB_NAME instance method must also deal with n:m tables (through tables).
The Index name creation on n:m tables uses the DB_NAME method to look up the table name to apply its index.
    
Defining the DB_NAME instance method in types.py on - at first - class DatasetTableSchema, set a placeholder for a the Amsterdam Schema look up in the near future.
For now, it uses an existing function that defines the table name which is also used in the baggob data processing.

### DIFF
--- a/schematools/contrib/django/factories.py
+++ b/schematools/contrib/django/factories.py
@@ -13,7 +13,6 @@ from schematools.types import (
     DatasetTableSchema,
     get_db_table_name,
 )
-from schematools import MAX_TABLE_LENGTH
 from schematools.utils import to_snake_case
 from .models import (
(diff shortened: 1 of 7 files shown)


jjmurre (approved):
Mooi dat je-em gevonden hebt!

vdboor (approved):

PR #105, opened 2021-02-16
Complete existing index creation (1:N relation reference)

cvriel:
The index creation logic was missing the creation of a databas index based on columns in a child table that refer to a parent table. The so called foreign key contraint is not enforced in the database. Yet the relevant columns are used in the API for retreiving the data. Also it can be used by database users who query on the tables directly.

### DIFF
--- a/schematools/importer/base.py
+++ b/schematools/importer/base.py
@@ -8,6 +8,7 @@ from typing import Dict, Optional
 from jsonpath_rw import parse
 from sqlalchemy import Column, ForeignKey, Index, Integer, MetaData, String, Table, exc, inspect
 
+from schematools import MAX_TABLE_LENGTH, TABLE_INDEX_POSTFIX
 from schematools.types import DatasetSchema, DatasetTableSchema
 from schematools.utils import to_snake_case
 
@@ -496,15 +497,16 @@ def index_factory(
 
     def make_hash_value(index_name):
         """
-        Postgres DB holds currently 63 max charakters for objectnames.
+        Postgres DB holds currently 63 max characters for objectnames.
         To prevent exceeds and collisions, the index names are shortend
         based upon a hash.
-        SHA1 holds a max output of 40 characters
+        With the blake2s algorithm a digest size is set to 20 bytes,
+        which proceduces a 40 character long heximal string plus
+        the additional 4 character postfix of '_idx' (TABLE_INDEX_POSTFIX).
         """
-        hash = hashlib.sha1()
-        hash.update(bytes(index_name, "utf-8"))
-        index_name = hash.hexdigest() + "_idx"
-        return index_name
+        return (
+            hashlib.blake2s(index_name.encode(), digest_size=20).hexdigest() + TABLE_INDEX_POSTFIX
+        )
 
     def define_fk_index():
         """creates an index on the columns that refer to another table as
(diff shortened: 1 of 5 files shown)


schematools/importer/base.py
```diff
@@ -478,19 +478,54 @@ def index_factory(
...
+    def make_hash_value(index_name):
+        """
+        Postgres DB holds currently 63 max charakters for objectnames.
+        To prevent exceeds and collisions, the index names are shortend
+        based upon a hash.
+        SHA1 holds a max output of 40 characters
+        """
+        hash = hashlib.sha1()
+        hash.update(bytes(index_name, "utf-8"))
+        index_name = hash.hexdigest() + "_idx"
```
gkoller, line 506: The `update`  is not really necessary. You could easily do this in one go:

```python
hashlib.sha1(index_name.encode().hexdigest()
```

Also using `encode()` is somewhat more idiomatic.

schematools/importer/base.py
```diff
@@ -478,19 +478,54 @@ def index_factory(
...
+    def define_fk_index():
+        """creates an index on the columns that refer to another table as
+        a child (foreign key reference)"""
+
+        if dataset_table.get_fk_fields():
+
+            for field in dataset_table.get_fk_fields():
+                field_name = f"{to_snake_case(field)}_id"
+                index_name = f"{db_table_name}_{field_name}_idx"
+                if len(index_name) > 64:
```
gkoller, line 518: You might want to use a constant for this value. Should this ever change in the default PostgreSQL installation or when basis_data custom compiles PostgreSQL we should be able to track this down easily. PostgreSQL calls this value `[NAMEDATALEN](https://www.postgresql.org/docs/11/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)`. Maybe we should do:

```python
PSQL_NAMEDATALEN = 64
```

This also allows you to to add an `assert` statement to the function `make_hash_value`, like:

```python
assert len(index_name) <= (
    PSQL_NAMEDATALEN - 1
), f"index names should be shorter than {PSQL_NAMEDATALEN}"
```
This guards against future Python developers that mindlessly change the hash function, because they've read somewhere that `sha1` is "insecure", change it to `sha256` or something else and suddenly we end up with larger index name than PostgreSQL can handle.

schematools/importer/base.py
gkoller: For what it is worth, I'm using [blake2s](https://docs.python.org/3.8/library/hashlib.html#hashlib.blake2s) in BBGA DAG. It allows you to specify a `digest_size` making it more explicit that you are after a hash of a specific size. In your code this is rather implicit in the choice of hash algorithm (`sha1`) though clearly documented in the docstring.

Anyway to generate a 40 character hash using blake2s you would do:

```python
hashlib.blake2s(index_name.encode(), digest_size=20).hexdigest()
```

Using a `digest_size` like that would obviate the need for the `assert` statement that I suggested in https://github.com/Amsterdam/schema-tools/pull/105/files#r577047793

gkoller (approved):

schematools/importer/base.py
cvriel: Verwerkt, bedankt voor deze tip

schematools/importer/base.py
cvriel: Goeie, er bestond al een constante voor object namen in Postgres. Die heb ik gebruikt. De assert heb ik gebruikt in een extra test op index naam.

PR #108, opened 2021-03-02
Guido/ds 331 enhance validator

gkoller:
Created a 'framework' for creating validators, together with one actual validator for checking that inferred PostgreSQL table names do not exceed 63 characters.

In addition, and as previously done for  Amsterdam/dataservices-airflow I have made the pre-commit hook and flake8 checks more strict. This generated so many warnings that I've dialed a few settings back.

The 'framework' and actual validator are all in the last commit.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,25 +1,45 @@
 [flake8]
-ignore=
-  # Black may add spaces in slice[func(v) : end] syntax
-  E203,
-  # Black leaves commas after combining lines
-  E231,
-  # Allow import * (for settings)
-  F403,
-  F405,
-  # Allow lambdas:
-  E731,
-  # line break before binary operator (incompatible with black):
-  W503,
-  # Allow raise Exception()
-  R102,
-exclude=.git,
-        **/migrations/*,
-        docs,
-        scripts,
-format = ${cyan}%(path)s${reset}:${yellow_bold}%(row)d${reset}:${green_bold}%(col)d${reset}: ${red_bold}%(code)s${reset} %(text)s
-# TODO: need to fix complexity, found in 9 functions
-#max-complexity=8
+# A = builtins
+# B = bugbear
+# C = comprehensions
+# D = docstrings
+# E = pycodestyle errors, rst
+# F = flake8 pyflakes, rst
+# G = logging-format
+# P = str-format
+# RST = rst-docstrings
+# S = bandit
+# T = print
+# W = pycodestyle warnings
+# B9 = bugbear opinions
+# ISC = implicit-str-concat
+ban-relative-imports = True
 max-line-length = 99
 statistics=True
-jobs=1
+# We should be using the following `select` line
+# but as that generates way too many warnings we
+# start with a smaller subset. However we should be gradually
+# moving to more and strict checks.
+# select = A, B, C, D, E, F, G, P, RST, S, T, W, B9, ISC
+select = B, C, D, E, F, G, RST, W, B9, ISC
+doctests = True
+ignore=
+    E203  # Black may add spaces in slice[func(v) : end] syntax
+    E231  # Black leaves commas after combining lines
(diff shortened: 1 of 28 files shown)


schematools/validation.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Registration is a side-effect of overriding this base class.
+
+    """
+
+    _registry: ClassVar[List[Type[Validator]]] = []
+    dataset: DatasetSchema
+
+    @classmethod
+    def __init_subclass__(cls: Type[Validator]) -> None:
```
jjmurre, line 94: Nice approach, this registration!

schematools/validation.py
```diff
@@ -0,0 +1,149 @@
...
+    supported by PostgreSQL.
+    """
+
+    def validate(self) -> Iterator[ValidationError]:
+        for table in self.dataset.get_tables(include_nested=True, include_through=True):
+            # `table_name` should probably be a property on `DatasetTableSchema`.
+            # There already is a `db_name` property however it truncates the inferred
+            # table name to prevent exceeding the PostgreSQL limit. Hence it cannot
+            # be used for validation purposes.
+            table_name = to_snake_case("_".join((self.dataset.id, table.id)))
```
jjmurre, line 142: In another PR I am adding a `name` property to the `DatasetTableSchema`. The `id` is provided by the amsterdam schema. The `name` is usually the same, but can be overridden in the amsterdam schema by a `shortname` property.
What this validation implementations does not cover yet are the tablenames that are generated on the fly for the through tables. That's for a new validator.

jjmurre (approved):
Looks good! It is easy now to add new validations. And furthermore, lots of improvements. In the spirit of the zen of python "types are a honking great idea -- let's do more of those!"  :-)

.flake8
```diff
@@ -1,25 +1,45 @@
...
+ban-relative-imports = True
 max-line-length = 99
 statistics=True
-jobs=1
+# We should be using the following `select` line
+# but as that generates way too many warnings we
+# start with a smaller subset. However we should be gradually
+# moving to more and strict checks.
+# select = A, B, C, D, E, F, G, P, RST, S, T, W, B9, ISC
+select = B, C, D, E, F, G, RST, W, B9, ISC
```
cvriel, line 24: Netjes

schematools/validation.py
gkoller: I thought I would get those through tables automatically by calling:

```python
get_tables(include_nested=True, include_through=True)
```

Are you saying `include_through=True` doesn't include through tables?

schematools/introspect/geojson.py
```diff
@@ -61,7 +61,7 @@ def geojson_to_table(geojson_features: Iterable[dict], file_name: str) -> List[d
 
     # Overlay all results into the amsterdam schem table format.
     result = []
-    for name, table_data in all_schema.items():
+    for name in all_schema.keys():
```
lbam, line 64: for name in all_schema:

schematools/validation.py
gkoller: Yeah, I saw a former co-worker use this once and thought "Does this really do what it seems to be doing?" Turns out it does: https://docs.python.org/3.8/reference/datamodel.html?highlight=__init_subclass__#object.__init_subclass__

New since 3.6. There's so much stuff in Python. I often feel I don't know half of it.

cvriel (approved):
Opzet voor validatie ziet er w.m.b. goed uit, lijkt erg handig

schematools/validation.py
gkoller: Will merge the PR. We can always extend the validator at a later time or write a new one as you suggested.

jjmurre: :facepalm: I totally missed the `include_through`. I'm glad this usecase is already covered!

PR #109, opened 2021-03-02
Jan/4390 shortname

jjmurre:
This PR is still a WIP. The airflow DAGs for GOB also need some refactoring to make the use of `shortname` possible.

More explanation is in the commit message.


ngaranko (approved):
LGTM

PR #111, opened 2021-03-08
Added bigint as datatype

cvriel:
Some sources provide us with an own id value. The value size of these source id's can exceed the maxinum of 32 bytes of the standard int(4) datatype in the database. During data load processing data from source into target table, the flow in Airflow can crash. Therefore the is a support added for bigint. Returing a 64 bytes int(8) datatype in the database.

### DIFF
--- a/schematools/importer/__init__.py
+++ b/schematools/importer/__init__.py
@@ -2,7 +2,7 @@ import numbers
 from decimal import Decimal
 
 from geoalchemy2 import Geometry
-from sqlalchemy import BigInteger, Boolean, Date, DateTime, Float, Integer, Numeric, String, Time
+from sqlalchemy import BigInteger, Boolean, Date, DateTime, Float, Numeric, String, Time
 from sqlalchemy.types import ARRAY
 
 from schematools.types import DatasetTableSchema
@@ -19,8 +19,7 @@ JSON_TYPE_TO_PG = {
     "string": String,
     "object": String,
     "boolean": Boolean,
-    "integer": Integer,
-    "biginteger": BigInteger,
+    "integer": BigInteger,
     "number": Float,
     "array": ARRAY(String),
     "https://schemas.data.amsterdam.nl/schema@v1.1.0#/definitions/id": String,
--- a/schematools/importer/__init__.py
+++ b/schematools/importer/__init__.py
@@ -2,7 +2,7 @@ import numbers
 from decimal import Decimal
 
 from geoalchemy2 import Geometry
-from sqlalchemy import Boolean, Date, DateTime, Float, Integer, Numeric, String, Time
+from sqlalchemy import BigInteger, Boolean, Date, DateTime, Float, Integer, Numeric, String, Time
 from sqlalchemy.types import ARRAY
 
 from schematools.types import DatasetTableSchema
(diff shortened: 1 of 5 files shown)


tests/test_import.py
```diff
@@ -96,3 +96,25 @@ def test_invalid_numeric_datatype_scale(here, engine, woningbouwplannen_schema,
...
+
+
+def test_biginteger_datatype(here, engine, woningbouwplannen_schema, dbsession):
+    """Prove that when biginter is used as datatype in schema, the datatype
+    in the database is set to datatype int(8) instead of int(4)"""
+    importer = BaseImporter(woningbouwplannen_schema, engine)
+    importer.generate_db_objects("woningbouwplan", ind_tables=True, ind_extra_index=True)
+    record = [
+        dict(r)
+        for r in engine.execute(
```
gkoller, line 108: `engine.execute()` returns a [ResultProxy](https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.ResultProxy). If you only expect one or only need the first row you can call `fetchone()` on the `ResultProxy.` The [RowProxy](https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.RowProxy) returned by `fetchone()` acts like a NamedTuple or dictionary. Meaning you don't need the list comprehension and the dict. In addition the `WHERE 1=1` is superfluous Combining everything you end up with:

```python
    results = engine.execute(
        """
        SELECT data_type, numeric_precision
        FROM information_schema.columns
        WHERE table_schema = 'public'
          AND table_name = 'woningbouwplannen_woningbouwplan'
          AND column_name = 'id_big';
    """
    )
    record = results.fetchone()
    assert record.data_type == "bigint"
    assert record.numeric_precision == 64
```
This is more succinct and expresses the intention arguably better.

schematools/importer/__init__.py
```diff
@@ -20,6 +20,7 @@
     "object": String,
     "boolean": Boolean,
     "integer": Integer,
+    "biginteger": BigInteger,
```
gkoller, line 23:
```
schema validate 'https://schemas.data.amsterdam.nl/schema@v1.1.1#' tests/files/woningbouwplannen.json
Structural validation: 
'biginteger' is not one of ['array', 'boolean', 'integer', 'null', 'number', 'object', 'string']

Failed validating 'enum' in schema[0]['properties']['type']:
    {'enum': ['array',
              'boolean',
              'integer',
              'null',
              'number',
              'object',
              'string']}

On instance['type']:
    'biginteger'
Reading schema from tests/files/woningbouwplannen.json
Semantic validation: success!
```

gkoller: I think we both misread the spec https://github.com/Amsterdam/amsterdam-schema/wiki/Amsterdam-Schema-Specification . Well at least I did.

The only valid JSON schema integral types are `integer` and `number`. The types mentioned in brackets after `integer`, eg:
```
integer (tiny8, small16, int32, bigint64)
```
are the types they map to depending on the values of minimum and maximum as you originally suggested. I'm afraid you did all this work due to me misreading it. I'm sorry

gkoller: What I'm saying is that according to JSON schema we can't have `biginteger` as a type on its own.

gkoller: Only realized that when I ran the validator and saw the error message.

jjmurre: See https://github.com/Amsterdam/amsterdam-schema/wiki/Grootst-Gemene-Deler  and search for `exclusiveMaximum` to see how integers were meant to be expressed.

tests/test_import.py
cvriel: :+1: Ah that much cleaner
The reason why I use where 1=1 is for the layout. The keep all the conditions with the same way (and ....) But where 1=1 can be removed because it has no use other then that

PR #114, opened 2021-03-11
Optimize toCamelCase

lbam:
Given

    >>> s = 'a_pretty_long_identifier_in_snake_case'

I get the following timeit results. Current implementation, with the @lru_cache decorator removed:

    >>> %timeit toCamelCase(s)
    10.4 µs ± 18.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

New version:

    >>> %timeit toCamelCase(s)
    4.88 µs ± 15.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

The LRU decorator is still there, since this is still a factor two below cache lookup speed:

    >>> %timeit toCamelCase(s)
    2.29 µs ± 21.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

### DIFF
--- a/schematools/utils.py
+++ b/schematools/utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from io import StringIO
 import json
 import re
 from typing import Dict, Final, Pattern
@@ -117,14 +118,28 @@ def schema_fetch_url_file(schema_url_file):
 
 
 @lru_cache(maxsize=500)  # type: ignore
-def toCamelCase(name):
+def toCamelCase(name: str) -> str:
     """
     Unify field/column/dataset name from Space separated/Snake Case/Camel case
     to camelCase.
     """
-    name = " ".join(name.split("_"))
-    words = RE_CAMEL_CASE.sub(r" \1", name).strip().lower().split(" ")
-    return "".join(w.lower() if i == 0 else w.title() for i, w in enumerate(words))
+    out = StringIO()
+    next_upper = False
+    first = True
+
+    for char in name:
+        if char == "_" or char.isspace():
+            next_upper = True
+            continue
+        if first:
+            char = char.lower()
+            first = False
+        elif next_upper:
+            char = char.upper()
+            next_upper = False
+        out.write(char)
+
+    return out.getvalue()
 
 
 @lru_cache(maxsize=500)  # type: ignore
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -12,6 +12,8 @@ def test_toCamelCase():
     assert toCamelCase("testNameMagic") == "testNameMagic"
     assert toCamelCase("TestNameMagic") == "testNameMagic"
     assert toCamelCase("test_name_magic") == "testNameMagic"
+    assert toCamelCase("numbers_33_in_the_middle_44") == "numbers33InTheMiddle44"
+    assert toCamelCase("per_jaar_per_m2") == "perJaarPerM2"
 
 
 def test_to_snake_case():


gkoller:
```

    def test_toCamelCase():
        """Confirm that:
        - space separated name is converted to camelCase
        - PascalCase results in camelCase
        - snake_case results in camelCase
        """
        assert toCamelCase("test name magic") == "testNameMagic"
        assert toCamelCase("test name magic2") == "testNameMagic2"
        assert toCamelCase("testNameMagic") == "testNameMagic"
>       assert toCamelCase("TestNameMagic") == "testNameMagic"
E       AssertionError: assert 'TestNameMagic' == 'testNameMagic'
E         - testNameMagic
E         ? ^
E         + TestNameMagic
E         ? ^

tests/test_utils.py:13: AssertionError
```

jjmurre (commented):
Ziet er leesbaar en elegant uit, nice! Mooie verbetering. Er faalt bij mij nog 1 testje, dat moet nog ff gefixed worden.

schematools/utils.py
```diff
@@ -117,14 +118,22 @@ def schema_fetch_url_file(schema_url_file):
...
     """
     Unify field/column/dataset name from Space separated/Snake Case/Camel case
     to camelCase.
     """
-    name = " ".join(name.split("_"))
-    words = RE_CAMEL_CASE.sub(r" \1", name).strip().lower().split(" ")
-    return "".join(w.lower() if i == 0 else w.title() for i, w in enumerate(words))
+    out = StringIO()
+    after_space = False
+    for c in s:
```
ngaranko, line 128: Single letter variables are really hard to read.
`s==value` and `c==character`?

Otherwise looks nice, though wonder how it will perform and if it will catch nasty edge cases we have (e.g. m2 and stuff).

lbam: I always use single-letter variable names for generic, short-lived stuff. I find value hard to parse :)

m2?

lbam:
Bug fixed at the expense of 0.4 µs per call.

schematools/utils.py
ngaranko: " generic, short-lived stuff." does not apply here, as `s` is effectively base value for function and `c` is accessed 4 times. ;)

m2 = https://github.com/Amsterdam/schema-tools/blob/master/tests/test_utils.py#L31
Would be nice to add test to these 2 cases:
```
    assert toCamelCase("numbers_33_in_the_middle_44") == "numbers33inTheMiddle44"
    assert toCamelCase("per_jaar_per_m2") == "perJaarPerM2"
```
note how `33in` appears over there.

lbam: I've changed s back to name. That was a leftover from an experimental version.

ngaranko: PS: `RE_CAMEL_CASE`, `re` and probably some of `typing` imports are not needed anymore.

schematools/utils.py
```diff
@@ -117,14 +118,28 @@ def schema_fetch_url_file(schema_url_file):
...
     """
     Unify field/column/dataset name from Space separated/Snake Case/Camel case
     to camelCase.
     """
-    name = " ".join(name.split("_"))
-    words = RE_CAMEL_CASE.sub(r" \1", name).strip().lower().split(" ")
-    return "".join(w.lower() if i == 0 else w.title() for i, w in enumerate(words))
+    out = StringIO()
+    after_space = False
+    first = True
```
ngaranko, line 128: `first` is not really needed:
```
after_space = True
for c in name.strip(" _"):
     ....
```
Will have same effect.

schematools/utils.py
lbam: I don't get why it should be numbers33inTheMiddle44. Why not numbers33InTheMiddle44? Surely "in" is a separate word.

schematools/utils.py
lbam: That's assuming spaces are all ASCII space (not newline, tab, NBSP). Can we make that assumption?

schematools/utils.py
ngaranko: Sorry for confusing, and you are totally correct. 
But will this `toCamelCase` generate `"numbers33InTheMiddle44"`? :)

schematools/utils.py
ngaranko: Nope. Valid point.

lbam: Also testNameMagic fails without first.

ngaranko: I admit, suggestion was not really good. I am just concerned that new version becomes not much more readable than previous ;)
Except it is not using horrible regex, which is plus, unless you are Perl developer.

schematools/utils.py
lbam: Solved.

gkoller:
On my computer this is even faster (3x)r:
```python
CAMEL_CASE_REPLACE_REGEXP: Final[Pattern[str]] = re.compile(r"(?:_|\s)+(.)")

def toCamelCase(name: str) -> str:
    """
    Unify field/column/dataset name from Space separated/Snake Case/Camel case
    to camelCase.
    """
    result = CAMEL_CASE_REPLACE_REGEXP.sub(lambda mo: mo.group(1).upper(), name)
    return result[0].lower() + result[1:]
```

lbam:
@gkoller I just discussed the new test with @ngaranko and we discovered that it was erroneous. It now checks for "numbers33InTheMiddle44" so your version is correct.

lbam:
3x speedup confirmed, and that regexp isn't too horrible. I can get .1µs off by separating out the lambda expression. That gets us very close to LRU lookup speed.

ngaranko (approved):
:clap:

gkoller:
However my version will leave `numbers33inTheMiddle44`, when given as input, with a lower cased `in` in the output. :-(

lbam:
@gkoller I'm merging this to get it over with for now, but please submit a PR with your version :)

ngaranko:
PS: there are unused vars left over.

PR #115, opened 2021-03-12
Guido/fix to camel case

gkoller:
Fix toCamelCase to titlecase words following numbers.

It previously didn't handle titlecasing words following numbers. Eg it didn't handle `fu33bar` correctly to become `fu33Bar`. Now it does.

In addition I changed the implementation to one based on a regex. A proof-of-concept regex version I had developed originally was 3x faster than the for loop. To handle additional cases I simply iterated on the proof-of-concept. It now is 3 µs slower that the for loop based one that does not handle the `fu33bar` case. If the for loop one starts handling that case it will likely loose its 3 µs edge.

In addition type annotated the entire module. Doing so surfaced a bug in function `schema_defs_from_url`, that is now fixed.

### DIFF
--- a/schematools/utils.py
+++ b/schematools/utils.py
@@ -2,13 +2,15 @@ from __future__ import annotations
 
 import json
 import re
-from typing import Dict, Final, Match, Pattern
+from pathlib import Path
+from typing import Any, Dict, Final, Match, Optional, Pattern, Type, Union, cast
 
 import requests
 from cachetools.func import lru_cache, ttl_cache
 from string_utils import slugify
 
-from . import MAX_TABLE_NAME_LENGTH, RELATION_INDICATOR, TMP_TABLE_POSTFIX, types
+from schematools import MAX_TABLE_NAME_LENGTH, RELATION_INDICATOR, TMP_TABLE_POSTFIX
+from schematools.types import ST, DatasetSchema, ProfileSchema
 
 RE_CAMEL_CASE: Final[Pattern[str]] = re.compile(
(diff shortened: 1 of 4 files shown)


schematools/utils.py
```diff
@@ -89,61 +92,98 @@ def def_from_url(base_url, data_type, dataset_name):
...
+
+
+CAMEL_CASE_REPLACE_PAT: Final[Pattern[str]] = re.compile(
+    r"""
+    (?:_|\s)+   # Find word boundaries by looking for underscore and whitespace characters, they
+                # will be discarded (not captured)
+    (.)         # Capture first letter of word on word boundary
+    |           # OR
+    (\d+)       # Capture a number
+    (?:_|\s)*   # Optionally followed by underscore an whitespace characters (to be discared)
```
lbam, line 130: Typo: an -> and

schematools/utils.py
```diff
@@ -89,61 +92,98 @@ def def_from_url(base_url, data_type, dataset_name):
...
     else:
         response = requests.get(schema_url_file)
         response.raise_for_status()
-        schema_location = response.json()
+        schema_data = response.json()
+
+    return cast(Dict[str, Any], schema_data)
+
+
+CAMEL_CASE_REPLACE_PAT: Final[Pattern[str]] = re.compile(
```
lbam, line 123: _CAMEL_CASE_REPLACE_PAT? We don't seem to be working with __all__ in this module.

lbam:
LGTM overall. Benchmark speed is 

    8 µs ± 94.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

here, which is still faster than what we started with. The RE is way more readable than RE_CAMEL_CASE and is well-explained.

lbam (approved):

PR #116, opened 2021-03-15
Rename DynamicModel.get_dataset into DynamicModel.get_dataset_schema for clarity

ngaranko:
Renaming confusing method name in order to give more clarity (and allow `get_dataset` method to be added in future).

### DIFF
--- a/schematools/contrib/django/models.py
+++ b/schematools/contrib/django/models.py
@@ -128,7 +128,7 @@ class DynamicModel(models.Model):
 
     @classmethod
     def get_dataset_schema(cls) -> DatasetSchema:
-        """Give access to the original dataset that this model is a part of."""
+        """Give access to the original dataset schema that this model is a part of."""
         return cls._table_schema._parent_schema
 
     @classmethod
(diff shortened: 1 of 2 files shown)


schematools/contrib/django/models.py
```diff
@@ -127,7 +127,7 @@ def __str__(self):
     # but this ensures the names don't conflict with fields from the schema.
 
     @classmethod
-    def get_dataset(cls) -> DatasetSchema:
+    def get_dataset_schema(cls) -> DatasetSchema:
         """Give access to the original dataset that this model is a part of."""
```
lbam, line 131: Maybe change the comment too?

vdboor: Scherp! :-)

vdboor (approved):

gkoller (approved):

schematools/contrib/django/models.py
ngaranko: Thanks. Fixed.

lbam (approved):

PR #117, opened 2021-03-15
Guido/fix to camel case empty string

gkoller:
* Better error reporting for `toCamelCase` (raise `ValueError` in case of empty string)
* Version bump

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ def read(filename: Union[Path, str]) -> str:
 
 setup(
     name="amsterdam-schema-tools",
-    version="0.17.10",
+    version="0.17.11",
     url="https://github.com/amsterdam/schema-tools",
     license="Mozilla Public 2.0",
     author="Amsterdam Data en Informatie",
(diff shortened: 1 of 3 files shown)


lbam:
What does to_snake_case do now? Because that's where the error came from in Amsterdam/dso-api#219.

gkoller:
> What does to_snake_case do now? Because that's where the error came from in [Amsterdam/dso-api#219](https://github.com/Amsterdam/dso-api/pull/219).

Oeps, just a sec.

setup.py
```diff
@@ -17,7 +17,7 @@ def read(filename: Union[Path, str]) -> str:
 
 setup(
     name="amsterdam-schema-tools",
-    version="0.17.9",
+    version="0.17.10",
```
ngaranko, line 20: Rebase, please :)

gkoller: Done


> What does to_snake_case do now? Because that's where the error came from in [Amsterdam/dso-api#219](https://github.com/Amsterdam/dso-api/pull/219).

raises a `ValueError` now as well.

lbam (approved):

PR #118, opened 2021-03-16
[4390] fix reload_schemas issue

jjmurre:
The shortnames possibily for tables created a subtle bug
when reloading schemas with the `reload_schemas` mgm command.

### DIFF
--- a/schematools/contrib/django/models.py
+++ b/schematools/contrib/django/models.py
@@ -259,7 +259,7 @@ class Dataset(models.Model):
             return
 
         new_definitions = {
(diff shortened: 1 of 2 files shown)


ngaranko (commented):
LGTM

ngaranko (approved):

lbam:
Is it possible to add a unit test?

jjmurre:
> Is it possible to add a unit test?

Good idea, I'll have a look. This code is only triggered from the DSO, so needs some Django plumbing to setup a test.

vdboor:
There is already `tests_django` for this, so all django plumbing is there.

PR #121, opened 2021-03-18
Nick/ad#5017

ngaranko:
This PR includes changes that will enable dataset versioning for Amsterdam Schema datasets.

`DatasetSchema` class (and JSON schema) extended with 3 additional properties:
 - `version` - comes from JSON schema definition parameter `version`, indicates version of given dataset
 - `default_version` - comes from JSON schema definition parameter `default_version`, contains default version of this dataset, should be set on all non-default versions;
 - `is_default_version` - calculated property, defaults to `True` for backwards compatibility. Indicates if given schema contains default version of dataset. (Calculated as `version == default_version`

Database Table creation process updated, when importing multiple versions of schema, table will be created for latest minor&patch version of each consecutive major version of dataset.

Default version of dataset (and all versions of dataset with matching major version) will have no major version number in table names.


mypy.ini
```diff
@@ -4,6 +4,7 @@ disallow_incomplete_defs = True
 disallow_untyped_calls = False
 disallow_untyped_decorators = True
 disallow_untyped_defs = True
+follow_imports = silent
```
cvriel, line 7: mindere strickte check voor nu :+1:

.pre-commit-config.yaml
```diff
@@ -29,11 +29,11 @@ repos:
...
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.812
-    hooks:
-      - id: mypy
-        language_version: python3
+  # - repo: https://github.com/pre-commit/mirrors-mypy
+  #   rev: v0.812
+  #   hooks:
+  #     - id: mypy
+  #       language_version: python3
```
vdboor, line 36: Should this be included?

note you can run `SKIP=mypy git commit` if you have to..

src/schematools/contrib/django/management/commands/create_tables.py
```diff
@@ -41,20 +42,30 @@ def create_tables(
...
     with connection.schema_editor() as schema_editor:
-        for model in models:
+        for db_table_name, models_group in models_by_table.items():
             # Only create tables if migration is allowed
             # - router allows it (not some external database)
             # - model is managed (not by default)
             # - user overrides this (e.g. developer)
-            db_table_name = model._meta.db_table
+            # - create table for latest version of this dataset group
+            model = list(reversed(sorted(models_group, key=lambda model: model._dataset.version)))[
```
cvriel, line 65: Meest recenste versie (hoogste versie nummer) wordt verwerkt als default begrijp ik

src/schematools/contrib/django/management/commands/import_schemas.py
```diff
@@ -62,16 +63,50 @@ def import_from_url(self, schema_url) -> List[Dataset]:
 
     def import_schema(self, name: str, schema: DatasetSchema) -> Optional[Dataset]:
         """Import a single dataset schema."""
+
+        created = False
         try:
-            dataset = Dataset.objects.get(name=to_snake_case(schema.id))
+            dataset = Dataset.objects.get(name=Dataset.name_from_schema(schema))
```
vdboor, line 69: Perhaps `Dataset.get_schema_name()` as classmethod name?

src/schematools/contrib/django/management/commands/import_schemas.py
```diff
@@ -62,16 +63,50 @@ def import_from_url(self, schema_url) -> List[Dataset]:
...
         except Dataset.DoesNotExist:
-            dataset = Dataset.create_for_schema(schema)
+            try:
+                # try getting default dataset by name and version
+                dataset = Dataset.objects.filter(Q(version=None) | Q(version=schema.version)).get(
+                    name=to_snake_case(schema.id)
```
vdboor, line 74: I guess this should also use `Dataset.name_from_schema()`, so better have it defined once in the method.

.pre-commit-config.yaml
gkoller: Agree with @vdboor 

And besides the `follow_imports = silent` should address most annoyances. (unless you're working on `types.py`).

ngaranko: Yes. I'll re-enable mypy in next pull request, which will include fixes for most (if not all) mypy errors in repo. Otherwise there's no reason to do skips during commit.

src/schematools/contrib/django/models.py
```diff
@@ -224,14 +226,26 @@ def __init__(self, *args, **kwargs):
             self.schema_data if "schema_data" in self.__dict__ and not self._state.adding else None
         )
 
+    @classmethod
+    def name_from_schema(cls, schema: DatasetSchema) -> str:
+        """Generate dataset name from schema"""
+        name = to_snake_case(schema.id)
+        if schema.version and not schema.is_default_version:
+            name = to_snake_case(f"{name}_{schema.version}")
```
vdboor, line 234: perhaps an `@` to concat name with version?

src/schematools/contrib/django/management/commands/create_tables.py
ngaranko: Only for database table creation, as Minor version should be fully backwards compatible.

src/schematools/contrib/django/management/commands/import_schemas.py
ngaranko: It is not really schema name, but rather Dataset name generator based on schema definition.

src/schematools/contrib/django/management/commands/import_schemas.py
```diff
@@ -62,16 +63,50 @@ def import_from_url(self, schema_url) -> List[Dataset]:
...
+            self.stdout.write(f"  Updated {name}")
+
+            if dataset.is_default_version != schema.is_default_version:
+                self.perform_version_update(dataset, schema)
+
+            dataset.save_for_schema(schema)
 
         return None
+
+    def perform_version_update(self, dataset: Dataset, schema: DatasetSchema) -> Dataset:
```
vdboor, line 93: nitpick: method name is a git confusing.. `update_dataset_version` ?

src/schematools/contrib/django/migrations/0010_auto_20210330_1447.py
```diff
@@ -0,0 +1,18 @@
...
+        ("datasets", "0009_auto_20210324_1157"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="datasettable",
+            name="db_table",
+            field=models.CharField(max_length=100),
+        ),
+    ]
```
vdboor, line 18: nitpick: You could place this in the other migration file, a bit silly to add 2 migration files in a single PR. preferably, the migration file has a clear name (`0009_add_versioning.py`)

src/schematools/contrib/django/models.py
ngaranko: Was planning to use it at first, but it will break wonderfully every time we use `to_snake_case`.

src/schematools/types.py
```diff
@@ -466,14 +482,33 @@ def is_nested_table(self):
         """Indicates if table is an nested table"""
         return self.nested_table
 
+    def model_name(self) -> str:
+        """Returns model name for this table. Including version number, if needed."""
+
+        from schematools.utils import to_snake_case
```
vdboor, line 488: does this need a local import?

src/schematools/contrib/django/migrations/0010_auto_20210330_1447.py
ngaranko: Oh, good one. Forgot to merge migrations.

src/schematools/contrib/django/management/commands/create_tables.py
gkoller: Untested:
```python
max(models_group, key=lambda model: model._dataset.version)
```
However both this version and your version will probably fail when we have two digit version number for either major, minor or path. Eg:
```python
print(next(reversed(sorted(("1.2.9", "1.1.5", "1.2.10")))))
```
will print `"1.2.9"` whereas it should have been `"1.2.10"`

tests/test_utils.py
```diff
@@ -3,7 +3,7 @@
 from schematools.utils import get_dataset_prefix_from_path, to_snake_case, toCamelCase
 
 
-def test_toCamelCase():
+def test_toCamelCase() -> None:
```
vdboor, line 6: 🤯  @gkoller  kunnen we dit uitzetten? helemaal voor de tests...

src/schematools/types.py
ngaranko: yes, avoiding circular imports. `schematools.utils` imports `schematools.types`

src/schematools/contrib/django/management/commands/create_tables.py
ngaranko: Updated.

tests/test_utils.py
gkoller: Yes and no. ;-)

Yes, to turn it off, add the following to `mypy.ini` or to `setup.cfg` after @ngaranko has migrated the mypy settings to `setup.cfg`:

```ini
[mypy-tests.*]
disallow_untyped_defs = False
disallow_incomplete_defs = False
disallow_untyped_decorators = False
```

No, as there is a school of thought, that I happen to adhere to, that states that your test code should follow the same quality standards as regular code. In case of Python code that would mean type annotated and documented (docstrings) test functions. After all, why should the test code be any different than regular code? Is it not as important? Don't you want to know the type of fixtures, so that you know what methods and attribute references can be made on them. Don't you want to know what exactly is being tested? I for sure want to know what kind of edge case or regression is being tested. So that at some point in time I can easily determine that the test is no longer required, or should definitely be kept.

I find that being as strict with test functions as with regular code leads to more meaningful tests and hence better test suites. The extra work required for a test to be written, means it will only be written if really serves a purpose. It cuts down on useless or no longer relevant tests.

Not saying that is the case here, but I'd rather see us getting into the habit. This is still a young project. Once we're a couple of years in, the extra effort will have paid of.

gkoller: For what it is worth I haven't even enabled all the flake8 warnings that I think should be turned on (eg docstrings for test functions)

src/schematools/contrib/django/models.py
vdboor: ah I checked and missed it: `from schematools import MAX_TABLE_NAME_LENGTH, RELATION_INDICATOR, TMP_TABLE_POSTFIX, types`

there is a ugly hack that might work:

```
from __future__ import annotations
import typing

...

# at bottom of the file:
if typing.TYPE_CHECKING:
    from schematools import  types
```

tests/test_utils.py
vdboor: Even if you want to have the same type strictness, mypy really won't like your pytest fixtures. will you be annotating each one of them? This increases test complexity.

(FYI, pycharm can resolve the fixture types if you set the test framework to "pytest" in your project)

src/schematools/types.py
vdboor: bummer. guess this can't befixded.

I completely missed it: `from schematools import MAX_TABLE_NAME_LENGTH, RELATION_INDICATOR, TMP_TABLE_POSTFIX, types`

FYI, there is a ugly hack that might work:

```
from __future__ import annotations
import typing

...

# at bottom of the file:
if typing.TYPE_CHECKING:
    from schematools import types
```

ngaranko: Yeah, but given other methods use same local import hack (and it's easier to read) - we'd do with local import here as well.

PR #125, opened 2021-03-23
Guido/ab5488 pre commit hook validate schema

gkoller:
:warning: **IMPORTANT** :warning: 
After this PR has been merged you need to do:
```
git pull --rebase=merges
pip uninstall amsterdam-schema-tools
find . -name '*.egg-info' -exec rm -rf {} \+
make install
```
Reason being that the directory structure has been changed significantly and editable pip installs might get "confused"

---

# This PR does four things:

- defines a new cli command `batch-validate`
- that is subsequently used in the definition of a `pre-commit` hook (eventually to be used in the `amsterdam-schema` repo)
- Mostly ditched `setup.py` in favor of a `setup.cfg` only config. What remains of `setup.py` is there specifically to allow editable pip installs.
- converted project to a `src` based layout.

In the process we added:
- Makefile targets (`clean`, `build`, `upload`)
- new project config `dev` (that what you supply to `pip install` in brackets `[dev]`)
- automatic project versioning and tagging

The `src` based layout changed to project from:
```
|-- docs
|-- schematools
|   |-- contrib
|   |-- events
|   |-- importer
|   |-- introspect
|   |-- maps
|   `-- provenance
|-- tests
|   `-- files
|-- tests_django
```
to:

```
|-- docs
|-- src
|   `-- schematools
|       |-- contrib
|       |-- events
|       |-- importer
|       |-- introspect
|       |-- maps
|       `-- provenance
|-- tests
|   |-- django
|   `-- files
```

The reason for doing so is that it:
- makes packaging easier. Simply point to `src` and you're done
- stricter separation between tests and the actual package 
- it is the generally recommended way these days to structure your Python project

Side effect is that we had to roll-up the the local `pytest` plugins (`fixtures.py`), that were solely used for capturing fixtures, into `conftest.py` files. Otherwise Django code was executed before we could configure Django.

The pre-`src` layout side stepped that issue by making `django_tests` part of the `schematools` package. Which it obviously is not: it is a collection of tests. But by doing so it was able to specify a `DJANGO_SETTINGS_MODULE`. With the new `src` layout, and tests that live entire out of the package, with had to utilize `pytest_configure` that dynamically configures Django before tests are executed.


Makefile
```diff
@@ -1,19 +1,33 @@
...
+clean:
+	find . -type d -name __pycache__ -exec rm -r {} \+
+	rm -rf build dist
+
+.PHONY: build
+build: clean
+	python -m build --sdist --wheel .
+
+.PHONY: upload
+	python -m twine upload --repository-url https://upload.pypi.org/legacy/ --username datapunt dist/*
```
vdboor, line 33: Hier mist de `target:`

gkoller: Good one. Fixing.

README.md
```diff
@@ -66,4 +66,28 @@ It is also possible to read GOB events from a batch file with line-separeted eve
...
+extend the `.pre-commit-config.yaml`
+in the project with the schema file defintions as follows:
+
+```yaml
+  - repo: https://github.com/Amsterdam/schema-tools
+    rev: v0.18.1
+    hooks:
+      - id: validate-schema
+        args: ['https://schemas.data.amsterdam.nl/schema@v1.1.1#']
+        exclude: 'datasets/index.json$'
```
jjmurre, line 85: When working on a branch of schema-tools, using an new dataset, still under development, this dataset is not yet at the production location (the `args`). Can this location be made configurable in some way?

setup.cfg
```diff
@@ -0,0 +1,75 @@
...
+    pg-grant
+    click
+    deepdiff
+    jsonschema
+    ndjson>=0.3.0
+    shapely
+    python-string-utils
+    python-dateutil
+    requests
+    jinja2
```
jjmurre, line 35: NB Chris has added some pinning here (security-related) in the `old` setup.py

setup.cfg
```diff
@@ -79,10 +79,16 @@ console_scripts =
...
+testpaths =
+    tests
+    tests_django
 DJANGO_SETTINGS_MODULE = tests_django.settings
 norecursedirs =
     .tox
     .git
+    dist
+    build
+    venv
```
jjmurre, line 94: Can we add `.venv` here?

vdboor (approved):

jjmurre (approved):
Nice improvement. Up to date with all the latest goodies!

README.md
gkoller: That's what the `args` command is for. You can change its value in `.pre-commit-config.yaml ` of the `amsterdam-schema` repo.

But the fact that we need it at all confuses me. Each schema already defines the meta schema it adheres to. Why then does the existing `validate` command, and hence the derived command `batch-validate`, require the meta schema to be explicitly specified?

setup.cfg
gkoller: will do

setup.cfg
gkoller: I'll look them up and will pin them here as well.

gkoller: I've have looked at [setup.py](https://github.com/Amsterdam/schema-tools/blob/master/setup.py) on `master` but I see no additional pinned versions. Are you referring to another project?

README.md
gkoller: Nevermind, because `jsonschema.validate` works like that.

Should we add some sort of preprocessing to it (in a different PR) so that it can select the meta schema from the schema, obviating the need for us to specify it explicitely?

setup.cfg
gkoller: Ah, this one https://github.com/Amsterdam/dso-api/pull/234

Will do the same here.

gkoller: On second thought, after having discussed it with @vdboor , I'll  leave the exact version pinning to the consumers of `schematools`. Eg, the applications that actually use this library. I will add `Django >= 3.0`, though. To state the minimum requirement.

tests/conftest.py
```diff
@@ -71,6 +111,8 @@ def dbsession(engine, dbsession, sqlalchemy_keep_db) -> Session:
     finally:
         # Drop all test tables after the tests completed
         if not sqlalchemy_keep_db:
+            from schematools.importer.base import metadata
+
             metadata.drop_all(bind=engine)
```
ngaranko, line 116: This `metadata` is undefined if `sqlalchemy_keep_db` is False.

ngaranko (approved):
LGTM. Only one minor fix needed.

setup.cfg
```diff
@@ -0,0 +1,152 @@
...
+    Programming Language :: Python
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+[options]
+package_dir =
+    = src
+packages = find:
+python_requires = >= 3.7
+install_requires =
+    sqlalchemy ~= 1.3.23
```
ngaranko, line 30: On other thought (and probably unrelated to this PR), looking at this massive requirements list makes me wonder if it would be nice to separate 'cli' requirements as well?
For example DSO api is not using CLI tools such as clock and pg-grant.

WDYT?

tests/conftest.py
gkoller: Good catch. Left over from an experiment that wasn't needed after all. Fixing.

setup.cfg
gkoller: Not sure. An optional requirements list is nice, but than the code should make provisions as well (catch import errors) and respond appropriately (whatever that may be) when optional dependencies are not present.

Taking it further in case  of `cli`. When that becomes optional, we should make the entry points (console apps) installation optional as well. Things get complicated rather quickly that way from a packaging point of view. In that case separating `cli` out of this project completely might be a better option.

`schematools` becomes `schematool-lib` (purely a library) and everything cli related would now become `schematools-cli` that has a dependency on `schematools-lib`. But that kind of a separation feels a little awkward.

So, I'm not sure. It definitely warrants more discussion.

README.md
gkoller: Resolving as the `args` parameter can be adjusted. But this definitely needs more discussion.

PR #127, opened 2021-03-25
Fix small docstring issues.

gkoller:
These became apparent after DSO API developere documentation
started including the `schematools` API docstrings.

These things are found earlier and easier once we start creating Sphinx
based documentation for this project as well.

Not bumping version number. These fixes will percolate to DSO API on the
next available PR.

### DIFF
--- a/src/schematools/validation.py
+++ b/src/schematools/validation.py
@@ -5,7 +5,7 @@ that allows us to perform semantic validation
 on a given JSON schema.
 
 Semantic validation is different from structural validation,
-as performed by the :module:`jsonschema`,
+as performed by the :mod:`jsonschema`,
 in that it takes into account the implied logic behind a JSON schema.
 For instance,
 we might indicate that what identifies a table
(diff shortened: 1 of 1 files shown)


src/schematools/validation.py
```diff
@@ -119,9 +119,9 @@ def run_all(self) -> Iterator[ValidationError]:
         """Run all registered validators.
 
         Yields:
-            :class:`ValidationError`s if any.
+            :class:`ValidationError`\s if any.
 
-        """
+        """  # noqa W605
```
ngaranko, line 124: This is incorrect syntax for `noqa` (and really common problem). This will hide ALL PEP8 errors:

Error code should be added after semicolon: `# noqa: W605`

---
Example:

```
x = """
     really long ass line to show that we have 2 errors in same block for real, no joke, best line ever
     :class:`ValidationError`\s if any.
x"""
```

Will result in:
```
test.py:4:80: E501 line too long (103 > 79 characters)
test.py:5:30: W605 invalid escape sequence '\s'
```

and with `# noqa: W605`

``` 
flake8 test.py       
test.py:4:80: E501 line too long (103 > 79 characters)
```

while `# noqa W605` will say file is completely correct.

gkoller: Excellent catch. Thank you. Will fix

jjmurre (approved):
Looks good, after fixing the `noqa` that Nick pointed out (good catch, I did'nt realize the importance of the colon).

PR #128, opened 2021-03-30
Fix dangling model references in the Django app registry

vdboor:


ngaranko (approved):

src/schematools/contrib/django/factories.py
```diff
@@ -21,6 +23,7 @@
 )
 
 TypeAndSignature = Tuple[Type[models.Field], tuple, Dict[str, Any]]
+MODEL_CREATION_COUNTER = 1
```
gkoller, line 26: Not sure how the dynamic models cache works (per process, per thread, something else), but this will almost certainly require thread/process synchronization (locking)

src/schematools/contrib/django/factories.py
```diff
@@ -321,6 +324,45 @@ def __call__(self, field: DatasetFieldSchema, dataset: DatasetSchema) -> TypeAnd
...
+        for model_name in dynamic_models:
+            del app_models[model_name]
+            del apps.app_configs[app_label].models[model_name]
+
+    # This also clears FK caches of other models that may have foreign keys:
+    apps.clear_cache()
+
+    # Allow code to detect whether a model is still alive despite having cleared all caches.
+    global MODEL_CREATION_COUNTER
+    MODEL_CREATION_COUNTER += 1
```
gkoller, line 358: Not sure how the dynamic models cache works (per process, per thread, something else), but this will almost certainly require thread/process synchronization (locking)

jjmurre (approved):
LGTM, maybe sprinkle in some extra type annotations in the new code.

src/schematools/contrib/django/app_config.py
```diff
@@ -0,0 +1,58 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class VirtualAppConfig(apps_config.AppConfig):
+    """
+    Virtual App Config, allowing to add models for datasets on the fly.
+    """
+
+    def __init__(self, apps, app_name, app_module):
```
gkoller, line 16: As we're trying to get things fully type annotated in `schematools` the members of this class could do with type annotations.

src/schematools/contrib/django/factories.py
```diff
@@ -321,6 +324,45 @@ def __call__(self, field: DatasetFieldSchema, dataset: DatasetSchema) -> TypeAnd
         return field_cls, args, kwargs
 
 
+def remove_dynamic_models():
```
gkoller, line 327: `-> None`

src/schematools/contrib/django/factories.py
```diff
@@ -321,6 +324,45 @@ def __call__(self, field: DatasetFieldSchema, dataset: DatasetSchema) -> TypeAnd
...
+
+    # This also clears FK caches of other models that may have foreign keys:
+    apps.clear_cache()
+
+    # Allow code to detect whether a model is still alive despite having cleared all caches.
+    global MODEL_CREATION_COUNTER
+    MODEL_CREATION_COUNTER += 1
+
+
+def is_dangling_model(model: Type[DynamicModel]):
```
gkoller, line 361: `-> bool`

src/schematools/contrib/django/factories.py
jjmurre: I assumed this is mainly added for testing? So one process/thread? Or is runtime behaviour also depending on this code?

src/schematools/contrib/django/management/commands/create_tables.py
```diff
@@ -26,7 +26,11 @@ def handle(self, *args, **options):
 
 
 def create_tables(
-    command: BaseCommand, datasets: Iterable[Dataset], allow_unmanaged=False, skip=None
+    command: BaseCommand,
+    datasets: Iterable[Dataset],
+    allow_unmanaged=False,
+    base_app_name=None,
+    skip=None,
 ):  # noqa:C901
```
gkoller, line 34: `-> None` ?

src/schematools/contrib/django/models.py
```diff
@@ -315,14 +316,14 @@ def schema_data_changed(self):
             and self.schema_data != self._old_schema_data
         )
 
-    def create_models(self) -> List[Type[DynamicModel]]:
+    def create_models(self, base_app_name=None) -> List[Type[DynamicModel]]:
```
gkoller, line 319: `base_app_name: Optional[str] = None`

src/schematools/contrib/django/factories.py
vdboor: good point. In the dso-api we have locking in place. I didn't dare to add locking at the low-level at this point but maybe we should?

lbam:
```py
class LockedCounter:
    """A thread-safe counter that starts at 1."""

    def __init__(self):
        self.__count = 0
        self.__lock = threading.Lock()

    @property
    def count(self):
        with self.__lock:
            return self.__count

    def next(self):
        with self.__lock:
            self.__count += 1
            return self.__count
```

vdboor: The problem is that the lock should cover a larger chunk of code: both the model creation and cache deletion. Thus, cache access should also be blocked, which a simple `threading.Lock` won't fix.

gkoller:
LGTM,

Just needs a:
```console
tbump 0.18.3
```
or, if you don't want it to push the version bump and tag commits automatically:
```console
tbump 0.18.3 --no-push
```

PR #131, opened 2021-04-06
Fix app_module on VirtualAppConfig, bug was visible in Django 3.1

vdboor:


jjmurre (approved):
Good catch!

ngaranko (approved):
Good one!

PR #133, opened 2021-04-07
Make pytest import django settings and fixtures only when needed.

ngaranko:
Make ugly setup even uglier, but hey, it works.

### DIFF
--- a/tests/django/conftest.py
+++ b/tests/django/conftest.py
@@ -1,179 +1,46 @@
 """Extra fixtures for ``schematools.contrib.django``"""
-import json
-
-import pytest
-
-from schematools.contrib.django.auth_backend import RequestProfile
-from schematools.contrib.django.models import Dataset, Profile
-from schematools.types import DatasetSchema
+from django.conf import settings
+from pytest_django.plugin import _setup_django
 
 
-@pytest.fixture
-def profile_medewerker():
-    return Profile.objects.create(
-        name="medewerker", scopes="['FP/MD']", schema_data={"datasets": {}}
-    )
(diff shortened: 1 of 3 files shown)


cvriel (approved):

tests/django/conftest.py
```diff
@@ -1,179 +1,46 @@
...
-def verblijfsobjecten_dataset(verblijfsobjecten_schema: DatasetSchema) -> Dataset:
-    """Create Verblijfsobjecten dataset."""
-    return Dataset.create_for_schema(verblijfsobjecten_schema)
-
-
-@pytest.fixture()
-def woningbouwplannen_dataset(woningbouwplannen_schema: DatasetSchema) -> Dataset:
-    """Create Woning Bouw Plannen dataset."""
-    return Dataset.create_for_schema(woningbouwplannen_schema)
+    config.pluginmanager.register(tests.django.fixtures)
```
gkoller, line 46: I struggled to get this to work after you pointed out that you could not get individual Django tests to run anymore after my restructuring changes to the `schematools` package.

So this is what it takes?
* explicit call to   `_setup_django()`
* and registering a new module containing the fixtures as a plugin

Very nice and straightforward! :+1: I learned something new!

gkoller (approved):

tests/django/conftest.py
ngaranko: Pretty much yes.

`_setup_django()` is actually called by `pytest-django` later on, but it would be too late for us.
And trick with module registration is needed because otherwise fixtures won't be recognized.w

PR #134, opened 2021-04-07
Replace deprecated django.contrib.postgres.fields.JSONField

lbam:
Booting up dso-api gives a warning about this one.


vdboor:
I think that also needs to be reflected in the migration files?

lbam:
Not sure, and the makemigration command complains about password authentication failing for user "bag_v11". What would break without it? Unit tests for schema-tools and dso-api run fine and I can still query datasets after autoreload.

lbam:
Also not sure if this is answer the question, but postgres.fields.JSONField is just models.JSONField with a deprecation warning added:

https://github.com/django/django/blob/main/django/contrib/postgres/fields/jsonb.py

lbam:
Got it, added the migration file.

vdboor (approved):

PR #135, opened 2021-04-12
Update pre-commit hook doc to exclude mete schema.

gkoller:
The pre-commit hook should not check the meta schema against itself.
Hence it requires an adjusted `exclude` argument.


cvriel (approved):

jjmurre (approved):
👍🏼

ngaranko (approved):

PR #143, opened 2021-04-28
[AB#6837] add DB schema name as parameter for table creation

cvriel:
Currently in Airflow an operator is used in serval dags to create the database objects (tables) that are defined in the dataschema definition (JSON schema). It defaults to the DB schema public. 

In some dags the tables are first created in a different schema then public before moved to public. Therefor it could be necessary to create the objects in a different DB schema then public. The DB schema name is added to specify the DB target schema. When not specified it defaults to public.

### DIFF
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -368,3 +368,73 @@ def test_size_of_index_name(engine, db_schema):
                 indexes_name.append(index["name"])
             for index_name in indexes_name:
                 assert len(index_name) <= (MAX_TABLE_NAME_LENGTH - len(TABLE_INDEX_POSTFIX))
+
+
+def test_index_creation_db_schema(engine, db_schema):
+    """Prove that identifier index is created based on DB schema specificiation """
+
+    test_data = {
+        "type": "dataset",
+        "id": "test",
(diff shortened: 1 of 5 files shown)


tests/test_indexes.py
```diff
@@ -368,3 +368,73 @@ def test_size_of_index_name(engine, db_schema):
...
+    for table in data["tables"]:
+        importer = BaseImporter(dataset_schema, engine)
+        # create DB schema
+        engine.execute("CREATE SCHEMA IF NOT EXISTS schema_foo_bar;")
+        # the generate_table and create index
+        importer.generate_db_objects(
+            table["id"], "schema_foo_bar", ind_tables=True, ind_extra_index=True
+        )
+
+        conn = create_engine(engine.url, client_encoding="UTF-8")
```
vdboor, line 427: Kan je hiervoor niet de `engine` fixture gebruiken?

Het hele idee van pytest fixtures is dat je minder setup code per test gaat schrijven. Kan ik je misschien nog een demo in geven.

cvriel: Ah ok check, ik zal er naar kijken!

PR #145, opened 2021-05-10
 Extra test and usage of existing db_table_name as a way to define temporary table name

cvriel:
- Extra test to check temporay table name when using existing db_table_name argument
    - Adjustment index factory to deal with db_table_name (which wasn't tested properly yet)


gkoller (approved):
LGTM

PR #153, opened 2021-05-30
Guido/ab4367 notify source owners failed import

gkoller:
Changed all `schematools.utils` function to also accept a `URL` in addition to a `str` for URLs.

While manually running MyPy on the module I noticed that there was a [typing issue](https://github.com/Amsterdam/schema-tools/pull/146/files/2a1fd8cccd622fb6621485e12da7d7b02bce79b9#r642109753) in it that could lead to an `AttributeError` at runtime. Puzzled as to why the MyPy pre-commit hook hadn't catched that, I noticed that hook had been disabled in commit 32a47f1fdf50618ea23820afee0f1b88d2f41a66. Though it wasn't document in that commit as to why it had been turned off (hint, hint), I happen remember it was done as it generated many errors and warning that were deemed an hindrance. However balancing the hindrance aspect compared to that of less actual errors in code I really prefer the latter. Hence I have re-enabled the MyPy commit hook. 

In addition I have added additional flake8 warnings (our code is not really consistent) and have to modify the `[flake8]` setion in `setup.cfg`. It turns out, after a lot of debugging, that flake8 doesn't like inline comments, though setuptools and pytest are seemingly fine with it.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+format = ${cyan}%(path)s${reset}:${yellow_bold}%(row)d${reset}:${green_bold}%(col)d${reset}: ${red_bold}%(code)s${reset} %(text)s
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,46 +2,57 @@
 # Supported hooks: https://pre-commit.com/hooks.html
 # Running "make format" fixes most issues for you
 repos:
+  - repo: https://github.com/pre-commit/mirrors-isort
+    rev: v5.8.0
+    hooks:
+      - id: isort
+        additional_dependencies: [toml]
   - repo: https://github.com/ambv/black
-    rev: 20.8b1
+    rev: 21.5b1
     hooks:
       - id: black
-        language_version: python3.8
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 3.9.2
     hooks:
       - id: flake8
         exclude: migrations/
         additional_dependencies:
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -2,46 +2,57 @@
...
-      - id: detect-private-key
-      - id: end-of-file-fixer
-      - id: trailing-whitespace
+      - id: check-json  # Checks json files for parseable syntax
+      - id: check-merge-conflict  # Check for files that contain merge conflict strings
+      - id: check-toml  # Checks toml files for parseable syntax
+      - id: check-yaml  # Checks yaml files for parseable syntax
+      - id: debug-statements  # Check for debugger imports and py37+ `breakpoint()` calls
+      - id: detect-private-key  # Detects the presence of private keys
+      - id: end-of-file-fixer  # Ensures that a file is either empty, or ends with one newline
```
cvriel, line 41: :+1: to the added explanation

cvriel (approved):

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+format = ${cyan}%(path)s${reset}:${yellow_bold}%(row)d${reset}:${green_bold}%(col)d${reset}: ${red_bold}%(code)s${reset} %(text)s
```
vdboor, line 2: Kan niet in `pyproject.toml`?

gkoller: No, flake8 does not support `pyproject.toml` and it can't be in `setup.cfg` (see commit message 53f69781da6831b9991b6ceed65c4390be0f51fd)

PR #155, opened 2021-06-09
Add --additional-schemas option to validate cli command

jjmurre:
Dataset with relations now also need the related datasets to be processed.

### DIFF
--- a/src/schematools/cli.py
+++ b/src/schematools/cli.py
@@ -228,7 +228,7 @@ def permissions_revoke(db_url, role):
 def permissions_apply(
     db_url,
     schema_url,
-    profile_url,  # NoQA
+    profile_url,
     schema_filename,
     profile_filename,
     pg_schema,
@@ -325,8 +325,11 @@ def _fetch_json(location: str) -> Dict[str, Any]:
 @schema.command()
 @option_schema_url
 @argument_schema_location
+@click.option("--additional-schemas", "-a", multiple=True)
 @click.argument("meta_schema_url")
-def validate(schema_url: str, schema_location: str, meta_schema_url: str) -> None:
+def validate(
+    schema_url: str, schema_location: str, additional_schemas: List[str], meta_schema_url: str
+) -> None:
     """Validate a JSON file against the amsterdam schema meta schema.
     schema_location can be a url or a filesystem path.
 
@@ -339,6 +342,9 @@ def validate(schema_url: str, schema_location: str, meta_schema_url: str) -> Non
 
     meta_schema = _fetch_json(meta_schema_url)
     dataset = _get_dataset_schema(schema_url, schema_location)
+
+    for schema in additional_schemas:
+        _get_dataset_schema(schema_url, schema)
 
     structural_errors = False
     try:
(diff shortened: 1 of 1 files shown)


src/schematools/cli.py
```diff
@@ -340,6 +343,9 @@ def validate(schema_url: str, schema_location: str, meta_schema_url: str) -> Non
     meta_schema = _fetch_json(meta_schema_url)
     dataset = _get_dataset_schema(schema_url, schema_location)
 
+    for schema in additional_schemas:
+        _get_dataset_schema(schema_url, schema)
```
gkoller, line 347: How does this work? `_get_dataset_schema` retrieves a schema and returns it. But the return value is discarded here. I don't see an assignment as happens in line 344.

src/schematools/cli.py
```diff
@@ -325,8 +325,11 @@ def _fetch_json(location: str) -> Dict[str, Any]:
 @schema.command()
 @option_schema_url
 @argument_schema_location
+@click.option("--additional-schemas", "-a", multiple=True)
```
gkoller, line 328: This parameter ` should either be documented with an extra `help` argument, or `validate`'s docstring should be amended with info about this new functionality/parameter

src/schematools/cli.py
jjmurre: Yes, this needs an explanation in a comment. When a dataset is loaded, it is also placed in a cache (side-effect, I know).
Actually, I want to get rid of manually loading related schemas and automatically preload schemas that are related. But that's for a next PR.

src/schematools/cli.py
jjmurre: Good point, I'll add a help argument.

gkoller (approved):

src/schematools/cli.py
gkoller: Ah, if it works like that then a comment would certainly be helpful. :-)

gkoller: BTW, I've had a brief look at the schema loading code. Can't seem to find where this side-effect is taking place. I would have expected a lock protected class variable that caches loaded instances, but there does not seem to be such a thing.

gkoller: Found it. They are all stored in a `schematools.datasetcollection.DatasetCollection`

Now I'm wondering, and definitely something for another PR:
* whether the metaclass for `DataCollection` should be a `ThreadSingleton` instead of a `Singleton` as it is currently. After all `schema-tools` is a package and used by the likes of `dso-api` and `dataservice-airflow` that both can be run in a threaded context
* whether we should even keep this solution (as you have been wondering yourself)
* should we make lookups dynamically/automatic preloading (instead of pre-caching them as we do here) we should rework the publishing of schema's. The name mangilng that happens currently upon publishing not only confused the hell out of me, but it apparently only done to adhere to some naming scheme common in the Javascript world. The problem with that name mangling is that the file system layout does not match the published object store layout. Making dynamic imports hard to implement when it needs to work in both environments.

src/schematools/cli.py
lbam: Do I understand correctly that the syntax would be

`--additional-schemas=foo --additional-schemas=bar`

? If so, I think the option name should be singular. Using a plural suggests that the argument itself is a list.

PR #157, opened 2021-06-14
AB#7272 temporal fields for nm relations

jjmurre:
This PR removes the initial implementation of event (to be used in a Kafka consumer).
GOB created a Kafka event producer, that was producing partial updates to events.
This led to a complex implementation for the consuming code and is not in line with Kafka best practices.

The second event producer is producing full events, and the associated Kafka consumer has been implemented,
so the initial implementation of the consumer could be ditched. 
Some extra complexity, automatic addition of temporal fields for relations, 
that was needed for the partial events, could also be ditched.

Temporal fields that are part of a relation are now also added and imported correctly. 
These fields are e.g. `beginGeldigheid` and `eindGeldigheid` 
and those fields are associated with a so called dimension (in this case the dimension `geldigOp`).

These dimension fields (that are part of a relation) have to be handled differently, they do not need to be prefixed with
the name of the relation field.


gkoller (approved):
I don't understand all the logic (haven't dealt with this part of the code yet), but the changes look good to me.

PR #158, opened 2021-06-14
Yrob/ab4388 temporaliteit naar tabel niveau

Yrob:
Nu temporaliteit in amsterdam-schema naar tabel niveau is verplaatst (https://github.com/Amsterdam/amsterdam-schema/pull/229). Moet dat ook hier gebeuren.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ exclude = '''
 github_url = "https://github.com/Amsterdam/schema-tools"
 
 [tool.tbump.version]
-current = "0.21.3"
+current = "0.21.4"
 regex = '''
   (?P<major>\d+)
   \.
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -490,14 +490,9 @@ class DatasetTableSchema(SchemaType):
         return self._parent_schema.use_dimension_fields
 
     @property
-    def temporal(
-        self, field_modifier: Optional[Callable] = None
-    ) -> Optional[Dict[str, Union[str, Dict[str, List[str]]]]]:
+    def temporal(self) -> Optional[Dict[str, Union[str, Dict[str, List[str]]]]]:
         """Return the temporal info"""
         from schematools.utils import to_snake_case
-
-        if field_modifier is None:
-            field_modifier = to_snake_case
 
         temporal_configuration = self["schema"].get("temporal", None)
         if temporal_configuration is None:
@@ -505,8 +500,8 @@ class DatasetTableSchema(SchemaType):
 
         for key, [start_field, end_field] in temporal_configuration.get("dimensions", {}).items():
             temporal_configuration["dimensions"][key] = [
-                field_modifier(start_field),
-                field_modifier(end_field),
+                to_snake_case(start_field),
+                to_snake_case(end_field),
             ]
 
         return temporal_configuration
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 
 import json
 from collections import UserDict
-from typing import Any, Callable, Dict, Iterator, List, NoReturn, Optional, Set, TypeVar, Union
+from dataclasses import dataclass, field
+from typing import Any, Dict, Iterator, List, NoReturn, Optional, Set, Tuple, TypeVar
 
 import jsonschema
 from methodtools import lru_cache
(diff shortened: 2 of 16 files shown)


src/schematools/types.py
```diff
@@ -527,14 +498,31 @@ def use_dimension_fields(self) -> bool:
         return self._parent_schema.use_dimension_fields
 
     @property
-    def temporal(self) -> Optional[Dict[str, Union[str, Dict[str, List[str]]]]]:
-        """Return the temporal info from the dataset schema """
-        return self._parent_schema.fetch_temporal(field_modifier=lambda x: x)
+    def temporal(
+        self, field_modifier: Optional[Callable] = None
```
gkoller, line 502: It would help if the `Callable` was more specific. What does the `Callable` accept as arguments and what does it return?

src/schematools/types.py
```diff
@@ -527,14 +498,31 @@ def use_dimension_fields(self) -> bool:
...
+    ) -> Optional[Dict[str, Union[str, Dict[str, List[str]]]]]:
+        """Return the temporal info"""
```
gkoller, line 504: You might want to document what `field_modifier` parameter is used for/actually does.

src/schematools/types.py
```diff
@@ -527,14 +498,31 @@ def use_dimension_fields(self) -> bool:
         return self._parent_schema.use_dimension_fields
 
     @property
-    def temporal(self) -> Optional[Dict[str, Union[str, Dict[str, List[str]]]]]:
-        """Return the temporal info from the dataset schema """
-        return self._parent_schema.fetch_temporal(field_modifier=lambda x: x)
+    def temporal(
+        self, field_modifier: Optional[Callable] = None
+    ) -> Optional[Dict[str, Union[str, Dict[str, List[str]]]]]:
```
gkoller, line 503: This was a pre-existing return type. I might have even introduced it. It was probably the best i could do at the time, not understanding any of the temporal logic. I still don't. But it looks horrendous and doesn't not help much in conveying what is actually returned here.

You seem to have a better understanding of the temporal logic, and are touching another dependent project (DSO-API). This is a good time to introduce a better data structure here instead of this amorphous dict.

What I am saying is that this seems a perfect moment to introduce a `dataclass` representing the temporal data in schema's. You should:
* define that `dataclass`, 
* provide proper docstrings, 
* and use it here to return it.

That should put us on the path to a better structured and typed Amsterdam Schema representating in Python.

@jjmurre What do you think?

jjmurre: +1 for a dataclass. The datastructure that needs to be represented is:
    
    temporal: {
      identifier: "volgnummer",
      dimensions: {
        geldigOp: [
          "beginGeldigheid",
          "eindGeldigheid"
       ]
    }
   }

So it would be quite straighforward to make this a dataclass.

jjmurre: Or actually two dataclasses I think, one separate dataclass for the inner `dimensions`.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ exclude = '''
 github_url = "https://github.com/Amsterdam/schema-tools"
 
 [tool.tbump.version]
-current = "0.21.3"
+current = "0.21.4"
```
vdboor, line 32: tbump zou ik niet in jouw branch doen. Liever pas bij het uitbrengen van een nieuwe release (na review en merge).
in DSO-API kan je met `pip install -e ../schema-tools` ook testen of alles werkt.

src/schematools/types.py
vdboor: hoe kan een property ook nog een parameter ontvangen?

src/schematools/types.py
vdboor: Definitely add a `TemporalSchema` class here, in line with the other classes (hence I wouldn't use a dataclass but schema dict like the rest). This also simplifies the DSO-API code. Instead of `.temporal.get('identifier', None)` you can just use `.temporal.identifier` as that's the actual field provided by this class.

src/schematools/types.py
vdboor: Nitpick: never call something "info" or "information", because it's meaningless. Everything in the computer memory is data or information. Better describe something like `"""Provides the configuration of the temporal table"""` or `"""Defines whether the table is temporal, and which fields are used to identify historical records."""`.

src/schematools/types.py
```diff
@@ -527,14 +498,31 @@ def use_dimension_fields(self) -> bool:
...
+        """Return the temporal info"""
+        from schematools.utils import to_snake_case
+
+        if field_modifier is None:
+            field_modifier = to_snake_case
```
vdboor, line 508: this is always true, because properties don't receive parameters.

src/schematools/types.py
```diff
@@ -910,7 +898,7 @@ def from_file(cls, filename: str):
 
     @classmethod
     def from_dict(cls, obj: Dict[str, Any]):
-        """ Parses given dict and validates the given schema """
+        """Parses given dict and validates the given schema"""
```
vdboor, line 901: When you rebase this branch on master (force push with lease is fine inside your own branch!), you'll find I've fixed all these black formatting errors.

src/schematools/types.py
jjmurre: Hier liep ik ook net tegenaan toen ik het lokaal testte. Vreemd eigenlijk dat python dit toestaat.

pyproject.toml
gkoller: Version bump met `tbump` doe ik altijd na de review, maar vóór de _rebase and merge_ naar `master`

src/schematools/types.py
gkoller: Damn, I missed this was defined as a property. Yes, a property can't have a (usable) parameter.

Yrob: Well I know someone else who didn't notice. :-) Still weird Python allows it. Can anyone think of a good reason for this? I removed the argument now. It doesnt seem to be needed anymore.

gkoller: Maybe someone had the following use case in mind:

```python
dataset: DatasetSchema = ...
...
temporal_meuk = DatasetSchema.fetch_temporal.fget(dataset, field_modifier=...)
```

jjmurre: That's exactly the usecase. The `field_modifier` was added to be backwards compatible. But, in hindsight, it would be better to remove it, it only complicates things and makes the code less readable.
The original implementation was doing an implicit `to_snake_case` on the result. Actually, it would be better to return the unaltered temporal info and leave it to the caller to do casing operations if needed.

src/schematools/types.py
Yrob: Would a dict of tuples be better for the dimensions? Or can a dimension have a different shape?

jjmurre: I would say temporality is always defined with a `start` and and `end`. So a 2-item list or a 2-item tuple. I agree with Diederik that, to keep things consistent, use a schemadict subclass.

Yrob: Bedoel je deze: `SchemaType(UserDict)`? Maar temporality is geen schema. En heeft dus geen id en type properties.

jjmurre (approved):
LGTM

PR #159, opened 2021-06-17
Fix types in typing.py

jjmurre:
Separate commit to fix the typings in types.py (and utils.py).

Changed some optional parameters (id, parent) for table and field to mandatory (non-keyword) parameters,
because the parameters are always needed anyway. This simplified the `mypy pleasing` quite a bit.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -4,7 +4,20 @@ from __future__ import annotations
 import json
 from collections import UserDict
 from dataclasses import dataclass, field
-from typing import Any, Dict, Iterator, List, NoReturn, Optional, Set, Tuple, TypeVar
+from typing import (
+    AbstractSet,
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    List,
+    NoReturn,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 import jsonschema
 from methodtools import lru_cache
@@ -24,11 +37,11 @@ class SchemaType(UserDict):
 
     @property
     def id(self) -> str:
-        return self["id"]
+        return str(self["id"])
 
     @property
     def type(self) -> str:
-        return self["type"]
+        return str(self["type"])
 
     def json(self) -> str:
         return json.dumps(self.data)
@@ -37,7 +50,7 @@ class SchemaType(UserDict):
         return self.data
 
     @classmethod
-    def from_dict(cls, obj: Dict[str, Any]) -> ST:
+    def from_dict(cls, obj: Dict[str, Any]) -> SchemaType:
         return cls(obj)
 
 
@@ -55,7 +68,7 @@ class DatasetSchema(SchemaType):
     This is a collection of JSON Schema's within a single file.
     """
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         """When initializing a datasets, a cache of related datasets
         can be added (at classlevel). Thus, we are able to get (temporal) info
         about the related datasets
@@ -65,7 +78,7 @@ class DatasetSchema(SchemaType):
         self.dataset_collection.add_dataset(self)
 
     @classmethod
-    def from_file(cls, filename: str):
+    def from_file(cls, filename: str) -> DatasetSchema:
         """Open an Amsterdam schema from a file."""
         with open(filename) as fh:
             return cls.from_dict(json.load(fh))
@@ -79,12 +92,12 @@ class DatasetSchema(SchemaType):
         return cls(obj)
 
     @property
-    def title(self) -> str:
+    def title(self) -> Optional[str]:
         """Title of the dataset (if set)"""
         return self.get("title")
 
     @property
-    def description(self) -> str:
+    def description(self) -> Optional[str]:
         """Description of the dataset (if set)"""
         return self.get("description")
 
@@ -99,38 +112,38 @@ class DatasetSchema(SchemaType):
         return self.get("url_prefix", "")
 
     @property
-    def identifier(self):
+    def identifier(self) -> str:
         """Which fields acts as identifier. (default is Django "pk" field)"""
         return self.get("identifier", "pk")
 
     @property
-    def version(self):
+    def version(self) -> str:
         """Dataset Schema Version"""
         return self.get("version", None)
 
     @property
-    def default_version(self):
+    def default_version(self) -> str:
         """Default version for this schema"""
         return self.get("default_version", self.version)
 
     @property
-    def is_default_version(self):
+    def is_default_version(self) -> bool:
         """Is this Default Dataset version.
         Defaults to True, in order to stay backwards compatible."""
         return self.default_version == self.version
 
     @property
-    def auth(self):
+    def auth(self) -> Optional[Union[str, List[str]]]:
         """Auth of the dataset (if set)"""
         return self.get("auth")
 
-    def get_dataset_schema(self, dataset_id) -> DatasetSchema:
+    def get_dataset_schema(self, dataset_id: str) -> DatasetSchema:
         return self.dataset_collection.get_dataset(dataset_id)
 
     @property
     def tables(self) -> List[DatasetTableSchema]:
         """Access the tables within the file"""
-        return [DatasetTableSchema(i, _parent_schema=self) for i in self["tables"]]
+        return [DatasetTableSchema(self, i) for i in self["tables"]]
 
     def get_tables(
         self,
@@ -145,7 +158,7 @@ class DatasetSchema(SchemaType):
             tables += self.through_tables
         return tables
 
-    @lru_cache()
+    @lru_cache()  # type: ignore[misc]
     def get_table_by_id(
         self, table_id: str, include_nested: bool = True, include_through: bool = True
     ) -> DatasetTableSchema:
@@ -214,11 +227,11 @@ class DatasetSchema(SchemaType):
         # we need to add a shortname to the dynamically generated
         # schema definition.
         if field.has_shortname or table.has_shortname:
-            snakecased_fieldname = to_snake_case(field.name)
+            snakecased_fieldname: str = to_snake_case(field.name)
             sub_table_schema["shortname"] = get_rel_table_identifier(
                 len(self.id) + 1, table.name, snakecased_fieldname
             )
-        return DatasetTableSchema(sub_table_schema, _parent_schema=self, nested_table=True)
+        return DatasetTableSchema(self, sub_table_schema, nested_table=True)
 
     def build_through_table(
         self, table: DatasetTableSchema, field: DatasetFieldSchema
@@ -277,7 +290,7 @@ class DatasetSchema(SchemaType):
         if relation is None and table.is_temporal:
             relation = field.relation
         right_dataset_id, right_table_id = [
-            to_snake_case(part) for part in relation.split(":")[:2]
+            to_snake_case(part) for part in str(relation).split(":")[:2]
         ]
 
         # XXX maybe not logical to snakecase the fieldname here.
@@ -286,7 +299,7 @@ class DatasetSchema(SchemaType):
         snakecased_field_id = to_snake_case(field.id)
         table_id = get_rel_table_identifier(len(self.id) + 1, table.id, snakecased_field_id)
 
-        sub_table_schema = {
+        sub_table_schema: Dict[str, Any] = {
             "id": table_id,
             "type": "table",
             "schema": {
@@ -318,20 +331,22 @@ class DatasetSchema(SchemaType):
 
         # Get the schema of the target table, to be able to get the
         # identifier fields.
-        target_table = table.dataset.get_dataset_schema(right_dataset_id).get_table_by_id(
-            right_table_id, include_nested=False, include_through=False
-        )
+        target_identifier_fields: Set[str] = set()
+        if table.dataset is not None:
+            target_table = table.dataset.get_dataset_schema(right_dataset_id).get_table_by_id(
+                right_table_id, include_nested=False, include_through=False
+            )
+            target_identifier_fields = set(target_table.identifier)
 
         # For both types of through tables (M2M and FK), we add extra fields
         # to the table (see docstring).
-        if field.is_through_table:
+        if field.is_through_table and target_identifier_fields:
             if field.is_object:
                 properties = field.get("properties", {})
             elif field.is_array_of_objects:
                 properties = field["items"].get("properties", {})
             else:
                 properties = {}
-            target_identifier_fields = set(target_table.identifier)
             # Prefix the fields for the target side of the relation
             extra_fields = {}
             for sub_field_id, sub_field in properties.items():
@@ -350,10 +365,10 @@ class DatasetSchema(SchemaType):
 
             sub_table_schema["schema"]["properties"].update(extra_fields)
 
-        return DatasetTableSchema(sub_table_schema, _parent_schema=self, through_table=True)
+        return DatasetTableSchema(self, sub_table_schema, through_table=True)
 
     @property
-    def related_dataset_schema_ids(self):
+    def related_dataset_schema_ids(self) -> Iterator[str]:
         """Fetch a list or related schema ids.
 
         When a dataset has relations,
@@ -393,7 +408,12 @@ class DatasetTableSchema(SchemaType):
     """
 
     def __init__(
-        self, *args, _parent_schema=None, nested_table=False, through_table=False, **kwargs
+        self,
+        _parent_schema: DatasetSchema,
+        *args: Any,
+        nested_table: bool = False,
+        through_table: bool = False,
+        **kwargs: Any,
     ) -> None:
         super().__init__(*args, **kwargs)
         self._parent_schema = _parent_schema
@@ -407,7 +427,7 @@ class DatasetTableSchema(SchemaType):
             raise ValueError("Invalid JSON-schema contents of table")
 
     @property
-    def name(self) -> Optional[str]:
+    def name(self) -> str:
         return self.get("shortname", self.id)
 
     @property
@@ -415,7 +435,7 @@ class DatasetTableSchema(SchemaType):
         return self.get("shortname") is not None
 
     @property
-    def dataset(self) -> Optional[DatasetSchema]:
+    def dataset(self) -> DatasetSchema:
         """The dataset that this table is part of."""
         return self._parent_schema
 
@@ -424,12 +444,10 @@ class DatasetTableSchema(SchemaType):
         """The description of the table as stated in the schema."""
         return self.get("description")
 
-    def get_fields(self, include_sub_fields=False) -> Iterator[DatasetFieldSchema]:
+    def get_fields(self, include_sub_fields: bool = False) -> Iterator[DatasetFieldSchema]:
         required = set(self["schema"]["required"])
         for id_, spec in self["schema"]["properties"].items():
-            field_schema = DatasetFieldSchema(
-                _id=id_, _parent_table=self, _required=(id_ in required), **spec
-            )
+            field_schema = DatasetFieldSchema(id_, self, _required=(id_ in required), **spec)
 
             # Add extra fields for relations of type object
             # These fields are added to identify the different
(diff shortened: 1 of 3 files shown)


src/schematools/types.py
```diff
@@ -99,38 +112,38 @@ def url_prefix(self) -> str:
...
 
     @property
-    def identifier(self):
+    def identifier(self) -> str:
         """Which fields acts as identifier. (default is Django "pk" field)"""
         return self.get("identifier", "pk")
 
     @property
-    def version(self):
+    def version(self) -> str:
```
vdboor, line 120:
```suggestion
    def version(self) -> Optional[str]:
```

src/schematools/types.py
```diff
@@ -99,38 +112,38 @@ def url_prefix(self) -> str:
...
     @property
-    def is_default_version(self):
+    def is_default_version(self) -> bool:
         """Is this Default Dataset version.
         Defaults to True, in order to stay backwards compatible."""
         return self.default_version == self.version
 
     @property
-    def auth(self):
+    def auth(self) -> Optional[Union[str, List[str]]]:
```
vdboor, line 136: laat maar weg, ik heb hier consistent `set[str]` van gemaakt in de auth refactor.

src/schematools/types.py
```diff
@@ -277,7 +290,7 @@ def build_through_table(
         if relation is None and table.is_temporal:
             relation = field.relation
         right_dataset_id, right_table_id = [
-            to_snake_case(part) for part in relation.split(":")[:2]
+            to_snake_case(part) for part in str(relation).split(":")[:2]
```
vdboor, line 293: Dit wordt dus `"None".split(":")[:2]` en geeft nog vreemdere situaties zo.

src/schematools/types.py
```diff
@@ -24,11 +37,11 @@ def __missing__(self, key: str) -> NoReturn:
 
     @property
     def id(self) -> str:
-        return self["id"]
+        return str(self["id"])
 
     @property
     def type(self) -> str:
-        return self["type"]
+        return str(self["type"])
```
vdboor, line 44: misschien liever `cast(str, self["type"])`?

src/schematools/types.py
```diff
@@ -214,11 +227,11 @@ def build_nested_table(
         # we need to add a shortname to the dynamically generated
         # schema definition.
         if field.has_shortname or table.has_shortname:
-            snakecased_fieldname = to_snake_case(field.name)
+            snakecased_fieldname: str = to_snake_case(field.name)
```
vdboor, line 230: is dat nodig? niet leiver `to_snake_case()` voorzien van een return type?

src/schematools/types.py
```diff
@@ -406,16 +426,25 @@ def __init__(
...
             raise ValueError("Invalid JSON-schema contents of table")
 
+    @classmethod
+    def from_dict(cls, obj: Dict[str, Any]) -> DatasetTableSchema:
+        """Create tableschema from a dict.
+
+        When this factory function is used, the _parent_schema argument is missing
+        and will be set to an empty DatasetSchema.
+        """
+        return cls(DatasetSchema({"id": "parent", "type": "dataset"}), obj)
```
vdboor, line 436: Is dat handig? straks kunnen we debuggen waarom er zo'n vreemde "parent" dataset rondspookt.

src/schematools/types.py
```diff
@@ -406,16 +426,25 @@ def __init__(
...
+    def name(self) -> str:
         return self.get("shortname", self.id)
 
     @property
     def has_shortname(self) -> bool:
         return self.get("shortname") is not None
 
     @property
-    def dataset(self) -> Optional[DatasetSchema]:
+    def dataset(self) -> DatasetSchema:
```
vdboor, line 447: moet dan ook weer anders

src/schematools/types.py
```diff
@@ -660,24 +688,35 @@ def __init__(
...
 
+    @classmethod
+    def from_dict(cls, obj: Dict[str, Any]) -> DatasetFieldSchema:
+        """Create fieldschema from a dict.
+
+        When this factory function is used, the _id and _parent_table arguments are missing
+        and will be set to the "id" of the obj argument and an empty DatasetSchema.
+        """
+        return cls(
+            obj["id"], DatasetTableSchema(DatasetSchema({"id": "parent", "type": "dataset"})), obj
```
vdboor, line 699: is dat handig, dat er een fake tabel in gezet wordt? Geeft ook weer andere rare fouten

src/schematools/types.py
gkoller: There's more to it. Let me look something up

gkoller: No, `version` is a required property according to the meta schema. Hence we should always be able to return a version. Meaning `-> str` is the correct return type

src/schematools/types.py
jjmurre: De id en parent waren optioneel, maar zijn eigenlijk overal nodig. Probleem als ze optioneel zijn is ook dat je een shitload extra checks nodig hebt om mypy tevreden te stellen.
Overigens wordt deze factory function voor de DatasetTable maar op 1 plek in de DSO gebruikt in een testje. We kunnen ook deze factory function weghalen. Die fake dataset zit mij idd ook niet lekker.

src/schematools/types.py
jjmurre: Goed punt, beter met een cast.

src/schematools/types.py
gkoller: The conversions or casts are necessary as MyPy has no idea what the types of the values associated with the keys `id` and `type` are. But I'm wondering if we can do better?

Schemas have a set a predefined attributes and a set of user definable attributes. If we would have had only predefined attributes we could have used a `dataclass` (or `TypedDict`). That way MyPy would have know that both `id` and `type` are of type`str`, obviating the need for casts/conversions.

However both `dataclass`es and `TypedDict`s do not support extra keys (the user definable attributes in the schemas, eg column names of tables). Hence we can't use a `dataclass` or even a `TypedDict` as the base class for our schema types. I'm wondering if there is anything out there that would allow for extra keys in Python data structures while still providing the type checking for the predefined keys?

src/schematools/types.py
```diff
@@ -37,7 +50,7 @@ def json_data(self) -> Dict[str, Any]:
         return self.data
 
     @classmethod
-    def from_dict(cls, obj: Dict[str, Any]) -> ST:
+    def from_dict(cls, obj: Dict[str, Any]) -> SchemaType:
```
gkoller, line 53: This is incorrect. `SchemaType.from_dict` is inherited by more specialized classes, eg `DatasetSchema`. If we call `DatasetSchema.from_dict` we do **not** want it to return a `SchemaType`, we want it to return a `DatasetSchema`. The `TypeVar` `ST` ensure exactly that.

gkoller: I'm sorry. Ignore my comment. `-> SchemaType` is correct.

src/schematools/types.py
gkoller: No, version is a required property according to the meta schema. Hence we should always be able to return a version. Meaning -> str is the correct return type

src/schematools/types.py
gkoller: `to_snake_case()` is already type annotated to return a `str` Hence MyPy should be able to derive this.

src/schematools/types.py
```diff
@@ -286,7 +299,7 @@ def build_through_table(
         snakecased_field_id = to_snake_case(field.id)
         table_id = get_rel_table_identifier(len(self.id) + 1, table.id, snakecased_field_id)
 
-        sub_table_schema = {
+        sub_table_schema: Dict[str, Any] = {
```
gkoller, line 302: This is a recurring datatype. Maybe introduce an alias for it?

```python
Json = Dict[str, Any]
```

Or to provide some top level type checking until MyPy support recursive types:

```python
Json = Union[Dict[str, Any], List, str, int, float, bool, None, object]
```

Yes `Json` instead of `JSON`. Why? Loved this explanation in https://nim-lang.org/docs/nep1.html

> In the age of HTTP, HTML, FTP, TCP, IP, UTF, WWW it is foolish to pretend these are somewhat special words requiring all uppercase. Instead treat them as what they are: Real words. So it's parseUrl rather than parseURL, checkHttpHeader instead of checkHTTPHeader etc.

src/schematools/types.py
```diff
@@ -475,12 +503,12 @@ def get_through_tables_by_id(self) -> List[DatasetTableSchema]:
         return tables
 
     @property
-    def display_field(self):
+    def display_field(self) -> Optional[Any]:
```
gkoller, line 506: Should this be `-> Optional[str]`

I am suggesting `str` as `display` is defined in the meta schema as:

```json
    "display": {
      "type": "string"
    },
```
However the docstring below suggest the result, if any, can be plural (fields). Should it be `Optional[List[str]]` then?

src/schematools/types.py
```diff
@@ -475,12 +503,12 @@ def get_through_tables_by_id(self) -> List[DatasetTableSchema]:
...
         """Tell which fields can be used as display field."""
         return self["schema"].get("display", None)
 
-    def get_dataset_schema(self, dataset_id):
-        """Return another datasets"""
+    def get_dataset_schema(self, dataset_id: str) -> Optional[DatasetSchema]:
+        """Return another dataset"""
         return self._parent_schema.get_dataset_schema(dataset_id)
```
gkoller, line 512: How does this function work? Every `DatasetSchema` has a `_parent_schema` of type DatasetSchema that is required (it is not `Optional`)

Calling `get_dataset_schema` on a `DatasetSchema` instance recursively calls `get_dataset_schema` on its parent. How does it stop? In other words what is its stopping condition?

src/schematools/types.py
```diff
@@ -536,43 +564,45 @@ def identifier(self) -> List[str]:
         # Convert identifier to a list, to be backwards compatible with older schemas
         if not isinstance(identifier, list):
             identifier = [identifier]
-        return identifier
+        return list(identifier)  # mypy pleaser
```
gkoller, line 567: Amazed this is needed to please MyPy. What version are you running?

And if it is really needed a `cast(List[str], ....)` would arguably be more appropriate.

src/schematools/types.py
```diff
@@ -536,43 +564,45 @@ def identifier(self) -> List[str]:
...
     @property
-    def filters(self):
+    def filters(self) -> Dict[str, Dict[str, str]]:
         """Fetch list of additional filters"""
-        return self["schema"].get("additionalFilters", {})
+        return dict(self["schema"].get("additionalFilters", {}))
 
     @property
-    def relations(self):
+    def relations(self) -> Dict[Any, Any]:
```
gkoller, line 595: I am not so familiar with relations. However `Any` for the key seems especially broad. I assume it can be `str` if I read the docstring correctly (_dictionary of names ..._)

Regarding the values. Does it resolve relations to actual table definitions? If so the return type can be `Dict[str, DatasetTableSchema]`.

If on the other hand it simple returns a JSON blob, you could turn it into `Dict[str, Json]` using the `Json` definition in https://github.com/Amsterdam/schema-tools/pull/159/files#r653344045

src/schematools/types.py
```diff
@@ -594,7 +624,7 @@ def model_name(self) -> str:
         model_name = self.id
         if self.dataset.version is not None and not self.dataset.is_default_version:
             model_name = f"{model_name}_{self.dataset.version}"
-        return to_snake_case(model_name)
+        return str(to_snake_case(model_name))  # mypy pleaser
```
gkoller, line 627: Should not be necessary. `to_snake_case` is already type annotated to return a `str`.

src/schematools/types.py
```diff
@@ -754,18 +793,18 @@ def provenance(self) -> Optional[str]:
...
     def get_dimension_fieldnames_for_relation(
-        self, relation: str, nm_relation: str
+        self, relation: Optional[str], nm_relation: Optional[str]
     ) -> Dict[str, List[str]]:
         """Gets the dimension fieldnames."""
         if relation is None and nm_relation is None:
             return {}
 
-        dataset_id, table_id = (relation or nm_relation).split(":")
+        dataset_id, table_id = str(relation or nm_relation).split(":")
```
gkoller, line 807: Hmm, so I presume MyPy can't determine, from the above if statement, that this or-expression evaluates to a `str`? Weird.

src/schematools/types.py
```diff
@@ -892,25 +931,25 @@ class ProfileSchema(SchemaType):
...
+    def name(self) -> Optional[str]:
         """Name of Profile (if set)"""
         return self.get("name")
 
     @property
-    def scopes(self):
+    def scopes(self) -> List[str]:
         """Scopes of Profile (if set)"""
-        return self.get("scopes")
+        return list(self.get("scopes", []))
```
gkoller, line 952: `cast(List[str], ...)`

src/schematools/utils.py
```diff
@@ -59,15 +67,22 @@ def profile_defs_from_url(profiles_url: Union[URL, str]) -> Dict[str, types.Prof
...
+        for p_name, p_schema in defs_from_url(
+            base_url=profiles_url, data_type=types.ProfileSchema
+        ).items()
+    }
 
 
-def defs_from_url(base_url: Union[URL, str], data_type: Type[types.ST]) -> Dict[str, types.ST]:
+def defs_from_url(
+    base_url: Union[URL, str], data_type: Type[types.SchemaType]
+) -> Dict[str, types.SchemaType]:
```
gkoller, line 80: You should revert this change. By returning `SchemaType` here you need all those explicit upcasts in `schema_def_from_url` and `schema_defs_from_url`. With the original `TypeVar` definition `ST` those upcasts (from `SchemaType` to more specific type`) will not be necessary anymore.

src/schematools/utils.py
```diff
@@ -59,15 +67,22 @@ def profile_defs_from_url(profiles_url: Union[URL, str]) -> Dict[str, types.Prof
...
     """Fetch all schema definitions from a remote file.
 
     The URL could be ``https://schemas.data.amsterdam.nl/datasets/``
     """
-    schema_lookup: Dict[str, types.ST] = {}
+    schema_lookup: Dict[str, types.SchemaType] = {}
```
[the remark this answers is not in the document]
jjmurre: Hm. Whenever used `ST` mypy was always complaining about it. Could this be a mypy version problem?

src/schematools/utils.py
```diff
@@ -109,7 +124,7 @@ def def_from_url(
         response = connection.get(base_url / index[dataset_id])
         response.raise_for_status()
 
-        dataset_schema: types.ST = data_type.from_dict(response.json())
+        dataset_schema: types.SchemaType = data_type.from_dict(response.json())
```
gkoller, line 127: `ST` was correct. By downcasting every possible `SchemaType` to `SchemaType` here, you'll need upcasts to more specific types elsewhere. `ST`'s definition prevents that.

src/schematools/types.py
jjmurre: That is not how it works. The `DatasetTableSchema` has a `_parent_schema` which is a `DatasetSchema`. So, there is no recursion involved.

src/schematools/types.py
jjmurre: My mypy version is 0.812 (feb 19, 2021), seems quite recent. I agree, cast is better.

src/schematools/types.py
gkoller: Pydantic seems to be able to provide the means to do exactly that with the following option:
```python
    class Config:
        extra = 'allow'
```

See also: https://pydantic-docs.helpmanual.io/usage/model_config/

* extra
  *  whether to ignore, allow, or forbid extra attributes during model initialization. Accepts the string values of 'ignore', 'allow', or 'forbid', or values of the Extra enum (default: Extra.ignore). 'forbid' will cause validation to fail if extra attributes are included, 'ignore' will silently ignore any extra attributes, and 'allow' will assign the attributes to the model.

src/schematools/types.py
jjmurre: Yes:
    returning Any from function declared to return "str"  [no-any-return]
Now solved it with a `cast`. I think you are right (other comment) that the return type of `to_snake_case`, because of the decorator, cannot be inferred correctly.

src/schematools/types.py
vdboor: Grappig, ik die ook altijd `checkHttpUrl` i.p.v. `checkHTTPURL`, maar als er niet geconcat wordt wel `URL` en dus `JSON` :-p. Hoewel ik wel weer een `QUrl` gebruik.

src/schematools/utils.py
```diff
@@ -59,7 +64,12 @@ def profile_defs_from_url(profiles_url: Union[URL, str]) -> Dict[str, types.Prof
 
     The URL could be ``https://schemas.data.amsterdam.nl/profiles/``
     """
-    return defs_from_url(base_url=profiles_url, data_type=types.ProfileSchema)
+    return {
+        p_name: p_schema
+        for p_name, p_schema in defs_from_url(
+            base_url=profiles_url, data_type=types.ProfileSchema
+        ).items()
+    }
```
vdboor, line 72: Hebben we dit echt nodig?

src/schematools/utils.py
```diff
@@ -36,7 +36,12 @@ def schema_defs_from_url(
         )
         return {dataset_name: schema}
 
-    return defs_from_url(base_url=schemas_url, data_type=types.DatasetSchema)
+    return {
+        ds_name: ds_schema
+        for ds_name, ds_schema in defs_from_url(
+            base_url=schemas_url, data_type=types.DatasetSchema
+        ).items()
+    }
```
vdboor, line 44: Hebben we dit echt nodig?

src/schematools/types.py
```diff
@@ -952,7 +998,12 @@ def tables(self) -> Dict[str, ProfileTableSchema]:
 class ProfileTableSchema(DatasetType):
     """A single table in the profile"""
 
-    def __init__(self, _id, _parent_schema=None, data: Optional[Dict[str, Any]] = None):
+    def __init__(
+        self,
+        _id: str,
+        _parent_schema: Optional[ProfileDatasetSchema] = None,
+        data: Optional[Dict[str, Any]] = None,
```
vdboor, line 1005: data niet gewoon verplicht maken i.p.v. optional?

src/schematools/types.py
```diff
@@ -923,7 +964,12 @@ def datasets(self) -> Dict[str, ProfileDatasetSchema]:
 class ProfileDatasetSchema(DatasetType):
     """A schema inside the profile dataset"""
 
-    def __init__(self, _id, _parent_schema=None, data: Optional[Dict[str, Any]] = None):
+    def __init__(
+        self,
+        _id: str,
+        _parent_schema: Optional[ProfileSchema] = None,
+        data: Optional[Dict[str, Any]] = None,
```
vdboor, line 971: Kunnen we data verplicht maken?

src/schematools/types.py
gkoller: Ah, thanks for the clarification. I had missed that this was `DatasetTableSchema`'s `get_dataset_schema`.

src/schematools/types.py
jjmurre: This fake parent was not a good idea. I reverted it and made the parameters optional again.

src/schematools/types.py
jjmurre: Restored to Optional. And added extra checks for None in the code where `dataset` is used.

src/schematools/types.py
jjmurre: Verwijderd.

src/schematools/utils.py
jjmurre: Kon gelukkig weer verwijderd worden agv. fix door @gkoller bij het gebruik van `ST` typevar.

src/schematools/utils.py
jjmurre: Niet meer, zie hierboven.

src/schematools/types.py
jjmurre: Zo te zien wel. Test blijven werken, ook DSO API tests blijven werken. Aangepast.

src/schematools/utils.py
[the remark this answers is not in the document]
jjmurre: You fixed it with an extra commit on the PR, tx!

src/schematools/utils.py
jjmurre: Fixed!

src/schematools/types.py
jjmurre: Using cast now.

src/schematools/types.py
jjmurre: Has always a fixed structure, changed it to `Dict[str, Dict[str, str]]`.

src/schematools/types.py
jjmurre: I tried the approach with the `Json` alias. The first (simple) variation works. However, for the `Union`  type it is not just introducing the alias and doing a search-replace. This generates 19 mypy new errors. So this Union needs some more investigation.

src/schematools/types.py
jjmurre: It should indeed be `Optional[str]`. Changed.

src/schematools/types.py
jjmurre: Added extra precaution, relation should never be None.

src/schematools/types.py
gkoller: You might consider making it an alias:
```python
Relations = Dict[str, Dict[str, str]]
...
def relations(self) -> Relations:
    ....
```
This way a future more strict (dataclass/Pydantic) based type can easily be introduced by search and replacing `Relations` with `<new-fancy-name-for-new-type>` and addressing the issues MyPy then finds.

gkoller (approved):
Excellent improvement! :+1:

PR #161, opened 2021-06-23
Yrob/#ab5522 separate id from url

Yrob:
API path en de ID property van datasets moeten onafhankelijk zijn om meerdere datasets met dezelfde naam te ondersteunen.
Daarom is aan de datasets het verplichte veld path toegevoegd.


src/schematools/contrib/django/models.py
```diff
@@ -238,12 +245,11 @@ def name_from_schema(cls, schema: DatasetSchema) -> str:
     def create_for_schema(cls, schema: DatasetSchema) -> Dataset:
         """Create the schema based on the Amsterdam Schema JSON input"""
         name = cls.name_from_schema(schema)
-
         return cls.objects.create(
             name=name,
             schema_data=schema.json_data(),
             auth=_serialize_claims(schema),
-            url_prefix=schema.url_prefix,
```
ngaranko, line 246: There's whole set of logic for `url_prefix` ;)
Like this one: https://github.com/Amsterdam/schema-tools/blob/0f499b3d422dfeaa66e90565c9e235efc9dcd288/src/schematools/utils.py#L83-L85

src/schematools/types.py
```diff
@@ -110,9 +110,9 @@ def license(self) -> Optional[str]:
         return self.get("license")
 
     @property
-    def url_prefix(self) -> str:
-        """Dataset URL prefix"""
-        return self.get("url_prefix", "")
+    def path(self) -> str:
+        """Dataset URL path"""
+        return self.get("path", "")
```
ngaranko, line 115: Would not it be better to set path default to `id`? Then we won't need extra mandatory parameter in schemas, that in 99% of cases is equal to ID.

Yrob: Wouldnt that be tricky with maintaining uniqueness? If some dataset_a.path == dataset_b.id

Yrob: I thought about parsing path from id. So id would be something like "haalcentraal__brk" and we could just convert that to "haalcentraal/brk" But then you might run into the 63 length limit for tables.

ngaranko: But that's same story as we have now with `path` specified on all schemas, isn't it? 

Migration in this PR uses `id` as `path` and in cases where Dataset needs to "cover" another id with url - developers can set `path` on one or both datasets in order to achieve desired result.

ngaranko: >I thought about parsing path from id. So id would be something like "haalcentraal__brk" and we could just convert that to "haalcentraal/brk" But then you might run into the 63 length limit for tables

In addition to 63 chars limits it will produce whole set of annoying snake_case to camelCase issues. :)

What I meant by path - is to make it default to id, but allow schemas to define path if needed. Hope this makes my comment a little bit more clear.

PR #164, opened 2021-06-30
Yrob/#ab5522 separate id from url

Yrob:
De datset ID moet worden losgekoppeld van het URL path. Zodat meerdere datasets met gelijke namen op verschillende paden mogelijk zijn. Api path wordt gebaseerd op het path binnen de objectstore.

De path property is toegevoegd aan het Dataset model. Het path wordt bij importeren afgeleid uit de path in de objectstore en toegevoegd aan de Dataset. Daarnaast worden  dataset tabellen in losse bestanden nu ondersteund door de importer, zodat versioned tabellen mogelijk blijven.


src/schematools/contrib/django/migrations/0011_auto_20210623_1135.py
```diff
@@ -0,0 +1,51 @@
...
+        migrations.RemoveField(
+            model_name="dataset",
+            name="url_prefix",
+        ),
+        migrations.AddField(
+            model_name="dataset",
+            name="path",
+            field=models.CharField(
+                null=True,
+                max_length=250,
```
gkoller, line 37: `path` is defined to be of type `string` with format `uri-reference` right? 

Can't seem to find the definition in `amsterdam-schema` now , but I did not recall a `maxLength` to be defined on that type. Hence there should **not** be one here either. I would suggest changing the field to `TextField` and **not** specifying a `max_length`

src/schematools/contrib/django/models.py
```diff
@@ -203,7 +208,9 @@ class Dataset(models.Model):
     enable_api = models.BooleanField(default=True)
     enable_db = models.BooleanField(default=True)
     endpoint_url = models.URLField(blank=True, null=True)
-    url_prefix = models.CharField(max_length=100, blank=True, validators=[URLPathValidator()])
+    path = models.CharField(
```
gkoller, line 211: `TextField`? See: https://github.com/Amsterdam/schema-tools/pull/164/files#r661758818

src/schematools/models.py
```diff
@@ -10,6 +10,7 @@
 class Dataset(Base):
     __tablename__ = "ams_schema_dataset"
     id = Column(String, primary_key=True)
+    path = Column(String)
```
gkoller, line 13: Here there is no length specified. This effectively boils down to the `TEXT` PostgreSQL type. So we now have two different definition for `path`:
- Django -> 250 -> `VARCHAR(250)
- SQLAlchemy -> unspecified -> `VARCHAR` -> `TEXT`

The SQLAlchemy one seems to match what I remember from the spec (when I could found it in the spec)

tests/conftest.py
```diff
@@ -132,13 +132,17 @@ def schemas_mock(requests_mock: Mocker, schema_url):
...
     # `requests_mock` is a fixture from the requests_mock package
-    afvalwegingen_json = HERE / "files" / "afvalwegingen.json"
-    requests_mock.get(
-        f"{schema_url}index.json", json={"afvalwegingen": "afvalwegingen/afvalwegingen"}
-    )
-    with open(afvalwegingen_json, "rb") as fh:
+    AFVALWEGINGEN_JSON = HERE / "files" / "afvalwegingen_sep_table.json"
+    CLUSTERS_JSON = HERE / "files" / "afvalwegingen_clusters-table.json"
+    requests_mock.get(f"{schema_url}index.json", json={"afvalwegingen": "afvalwegingen"})
+    with open(AFVALWEGINGEN_JSON, "rb") as fh:
```
gkoller, line 138: The constant `AFVALWEGINGEN_JSON` suggests it concerns a JSON file. JSON is text-based. The file is opened in binary though. Why?

tests/conftest.py
```diff
@@ -132,13 +132,17 @@ def schemas_mock(requests_mock: Mocker, schema_url):
...
         requests_mock.get(
-            f"{schema_url}afvalwegingen/afvalwegingen",
+            f"{schema_url}afvalwegingen/dataset",
+            content=fh.read(),
+        )
+    with open(CLUSTERS_JSON, "rb") as fh:
```
gkoller, line 143: See issue as in https://github.com/Amsterdam/schema-tools/pull/164/files#r661767350

tests/conftest.py
```diff
@@ -132,13 +132,17 @@ def schemas_mock(requests_mock: Mocker, schema_url):
...
     """
     # `requests_mock` is a fixture from the requests_mock package
-    afvalwegingen_json = HERE / "files" / "afvalwegingen.json"
-    requests_mock.get(
-        f"{schema_url}index.json", json={"afvalwegingen": "afvalwegingen/afvalwegingen"}
-    )
-    with open(afvalwegingen_json, "rb") as fh:
+    AFVALWEGINGEN_JSON = HERE / "files" / "afvalwegingen_sep_table.json"
+    CLUSTERS_JSON = HERE / "files" / "afvalwegingen_clusters-table.json"
+    requests_mock.get(f"{schema_url}index.json", json={"afvalwegingen": "afvalwegingen"})
```
gkoller, line 137: Change `conftest.schema_url` to return an `URL` and then this can become:
```python
requests_mock.get(schema_url / "index.json", ....
```

tests/conftest.py
```diff
@@ -132,13 +132,17 @@ def schemas_mock(requests_mock: Mocker, schema_url):
...
+    with open(AFVALWEGINGEN_JSON, "rb") as fh:
         requests_mock.get(
-            f"{schema_url}afvalwegingen/afvalwegingen",
+            f"{schema_url}afvalwegingen/dataset",
```
gkoller, line 140: See also https://github.com/Amsterdam/schema-tools/pull/164/files#r661769631

```python
schema_url / "afvalwegingen/dataset"
```

tests/conftest.py
Yrob: Requests_mock.get slikt alleen binary data.

src/schematools/contrib/django/migrations/0011_auto_20210623_1135.py
Yrob: its no longer defined in amsterdam-schema, but equal to the path in the objectstore. This is a relic from the earlier version where it was defined in the schema. Forgot to change it. Good catch

tests/conftest.py
Yrob: good one, will do.

PR #165, opened 2021-07-02
Fix the Django classname for a "through" model

jjmurre:
The through classname that is automatically generated
should use `field.id` and not `field.name`,
because, when a `shortname` is defined for a field
in the amsterdam schema, this should not affect
the name of the generated through class.

Because, for table fields in amsterdam schema:

`field.id` is always the id in the schema
`field.name` is also the id in the schema,
unless there is a `shortname` defined, in that case
`field.name` is the `shortname`.


gkoller:
How does this effect, if at all, the length of the DB table names?

jjmurre:
> How does this effect, if at all, the length of the DB table names?

Tables and fields have a `db_name` methods that is used to determine the name to be used in postgres. This`db_name` uses the `table.name` and `field.name` an thus will use a `shortname` if that is defined.

ngaranko (approved):

Yrob (approved):

PR #167, opened 2021-07-08
Paths set correctly when importing

Yrob:
Bugfix. Dataset paths were not set properly when importing schemas.
This update fixes that.


ngaranko (approved):
Test me please :D

PR #170, opened 2021-07-12
AB#4378 Add id column for through tables

jjmurre:
The temporal fields for "geldigheid" need to be taken from
the through table. However, to be able to query for these fields
in DSO API, this through table needs to have a non-compound primary key.

This PR adds such a PK field and adjusts the imports to give this
PK field the value of the concatenation of both sides of the relation.

### DIFF
--- a/src/schematools/events/full.py
+++ b/src/schematools/events/full.py
@@ -160,12 +160,15 @@ class DataSplitter:
                 nm_row_data[f"{name_prefix}{fn}"] = fv
 
             # id for source side of relation
-            nm_row_data.update({snaked_source_table_id: id_value})
+            nm_row_data[snaked_source_table_id] = id_value
 
             # id for target side of relation
-            nm_row_data[f"{snaked_field_id}_id"] = ".".join(
-                (str(row_data[fn]) for fn in target_identifier_fields)
-            )
+            target_id_value = ".".join((str(row_data[fn]) for fn in target_identifier_fields))
+            nm_row_data[f"{snaked_field_id}_id"] = target_id_value
+
+            # PK field, needed by Django
+            nm_row_data["id"] = f"{id_value}{target_id_value}"
+
             nm_rows.append(nm_row_data)
 
         self.junction_table_rows[snaked_field_id] = nm_rows
--- a/src/schematools/importer/base.py
+++ b/src/schematools/importer/base.py
@@ -7,7 +7,18 @@ from typing import Dict, Optional
 
 from jsonpath_rw import parse
 from more_itertools import first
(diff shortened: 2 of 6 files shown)


src/schematools/events/full.py
```diff
@@ -163,9 +163,12 @@ def _handle_junction_table(self, field_id: str, field_data: dict, nm_relation):
...
 
             # id for target side of relation
-            nm_row_data[f"{snaked_field_id}_id"] = ".".join(
-                (str(row_data[fn]) for fn in target_identifier_fields)
-            )
+            target_id_value = ".".join((str(row_data[fn]) for fn in target_identifier_fields))
+            nm_row_data[f"{snaked_field_id}_id"] = target_id_value
+
+            # PK field, needed by Django
+            nm_row_data.update({"id": f"{id_value}{target_id_value}"})
```
gkoller, line 170: I would suggest either

```python
nm_row_data.update(
    {
        f"{snaked_field_id}_id": target_id_value,
        "id": f"{id_value}{target_id_value}"
    }
)
```
or
```python
nm_row_data[f"{snaked_field_id}_id"] = target_id_value
nm_row_data["id"] = f"{id_value}{target_id_value}"
```

src/schematools/importer/base.py
```diff
@@ -437,6 +437,10 @@ def table_factory(
 
                     # We need a 'through' table for the n-m relation
                     sub_columns = [
+                        Column(
+                            "id",
+                            String,
```
gkoller, line 442: Not a fan of `String` without a `length`. String maps to `VARCHAR`. In most databases that requires a length to be specified. If we don't care about the length we should use `Text`, which maps to `TEXT`: the canonical PostgreSQL type for textual data without a length.

* https://docs.sqlalchemy.org/en/13/core/type_basics.html?highlight=string#sqlalchemy.types.String
* https://docs.sqlalchemy.org/en/13/core/type_basics.html?highlight=string#sqlalchemy.types.Text

BTW I do realize all existing code uses `String` instead of `Text`. Just wanted to have it pointed out once.

jjmurre: I think in this case I'll use a fixed length varchar. This column also needs an index (reminds me, I have to check on that). Using `TEXT` does not feel right here.

src/schematools/events/full.py
jjmurre: Totally agree, this is a weird mix.

src/schematools/importer/base.py
jjmurre: Hm. On second though, PostgreSQL docs say:

    VARCHAR (without the length specifier) and TEXT are equivalent.

I'll change it to `Text`, that make the intent most clear. As far as I can see, there is no problem with adding an index to `Text` type columns.

PR #172, opened 2021-07-13
Make the `id` column for through table a PK

jjmurre:
An extra singular `id` field has been added as PK to the through tables. For efficiency, an index need


gkoller (approved):
Look good to me.

PR #173, opened 2021-07-15
Fix: check max ident length for indices.

gkoller:
The check was done for foreign keys, but not for indices. This corrects
that.


jjmurre (approved):
👍🏼

PR #174, opened 2021-07-19
[AB#6554] Decouple UserScopes from is_authorized_for

lbam:
Just passing the scopes to the `UserScopes` constructor makes the code easier to work with (to debug especially). It also means that we can introduce more sensible logging than the per call logging done by `authorization_django`, especially in `UserScopes.has_any_scope`, which can do many `is_authorized_for` calls.

### DIFF
--- a/src/schematools/permissions/auth.py
+++ b/src/schematools/permissions/auth.py
@@ -21,22 +21,7 @@ from schematools.types import (
     ProfileTableSchema,
 )
 
-HasAllScopesFunc = Callable[..., bool]  # can't describe func(*scopes: str) -> bool
-
-__all__ = (
-    "UserScopes",
-    "create_scopes_check",
-)
-
-
-def create_scopes_check(*user_scopes: str) -> HasAllScopesFunc:
-    """Create a function that tests whether the scopes are set.
-
-    This function only needs to be used when features such as
-    ``request.is_authorized_for(*scopes)`` from authorization_django are not available.
-    """
-    user_scopes_set = set(user_scopes)  # can be empty
-    return lambda *needed: user_scopes_set.issuperset(needed)
+__all__ = ("UserScopes",)
 
 
 def abort_on_highest(func: Callable[..., Permission]):
@@ -90,7 +75,9 @@ class PermissionCollection:
 class UserScopes:
     """A request-like object that tells what the current user may access.
 
-    This is the foundation for all permission checks.
+    A UserScopes encapsulates the scopes on a request and the profiles that apply,
+    and performs permission checks against a schema.
+
     All ``has_...()`` functions are used for permission checks.
     Internally, these read the schema and profile data for the authorization matrix.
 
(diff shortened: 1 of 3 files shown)


vdboor:
Fine by me. Note this does mean we also no longer support the middleware settings in authorization_django (ALWAYS_OK and FORCED_ANONYMOUS_ROUTES) but I doubt we'd use them anyway..

lbam:
`ALWAYS_OK` is something we can easily emulate. I don't know what the use case for `FORCED_ANONYMOUS_ROUTES` is, because it actually requires us to [never call](https://github.com/Amsterdam/authorization_django/blob/97194d7a61deb25ac30ad2954bc913cc6ec28887/authorization_django/middleware.py#L69) `is_authorized_for`.

I'll prepare a PR for dso-api.

src/schematools/permissions/auth.py
```diff
@@ -323,6 +312,12 @@ def _may_include_profile_table(self, profile_table: ProfileTableSchema):
             _match_filter_rule(rule, self._query_param_names) for rule in mandatory_filtersets
         )
 
+    def __str__(self) -> str:
+        """str implementation, for debugging purposes."""
+        self._materialize_profiles()
```
vdboor, line 317: deze call is hier toch niet nodig?

lbam: Oh, ik had profiles en scopes door elkaar gehaald. Brainfart, ik haal het weg.

lbam:
@vdboor Nog commentaar?

jjmurre (approved):
LGTM.

src/schematools/contrib/django/migrations/0012_schema_data_as_textfield.py
```diff
@@ -1,23 +0,0 @@
-# Generated by Django 3.1.12 on 2021-07-20 11:09
```
ngaranko, line 1: Deleting migrations is rather bad practice and should be performed with great care.
Assuming both deleted migrations are applied to acceptance and production: we would have to manually revert them and then push current branch into master, otherwise we're risking deployment crash next time new migration is added to schematools.

src/schematools/permissions/auth.py
```diff
@@ -288,7 +274,7 @@ def get_active_profile_datasets(self, dataset_id: str) -> List[ProfileDatasetSch
         This already checks whether the mandatory user scopes are set.
         """
         if self._all_profiles is None:
-            raise RuntimeError("load_profiles() is not called yet")
+            self._all_profiles = None
```
ngaranko, line 277: It might be bad idea to remove `RuntimeError` here, old behavior was intentionally breaking when profiles are not loaded yet, avoiding issues with authorization being checked before profiles are ready to use.

New code (not only it is redundant) hides this issue, allowing Profiles to be `None` (not loaded, because Middleware would provide empty list if profiles are not in system).
Or you're planning to change this as well?

ngaranko:
Could you please describe reasoning behind this change?

`is_authorized_for` function that is used in almost all python projects within our codebase, and even though it's not very intelligent one and was easily replaced by this PR - it does more than just scopes `issubset` check, which means shared functionality would not be available in schematools.

----

Great idea to finally start using `JSONField`! Not really sure why anymore, but we could not use it before.

src/schematools/contrib/django/migrations/0012_schema_data_as_textfield.py
lbam: This was a temporary workaround for an issue that is now fixed. Removing the commit that did this.

lbam:
The only other thing `is_authorized_for` does is logging, and it's log message only includes the scopes being checked, not the URL path or other details of the request. It's also harder to reason about/debug than a simple check of two sets of scopes.

src/schematools/permissions/auth.py
lbam: Changed this to `self._all_profiles = []`. This is needed to make the tests in dso-api work. Discussed this with @vdboor on Slack.

ngaranko:
> The only other thing is_authorized_for does is logging, and it's log message only includes the scopes being checked, not the URL path or other details of the request. It's also harder to reason about/debug than a simple check of two sets of scopes.

@lbam: it is not really that simple, just like @vdboor mentioned, we're loosing part of functionality related to `ALWAYS_OK` and `FORCED_ANONYMOUS_ROUTES`, together with any possible future improvements.
This change is not about what code does, but rather about functionality that is shared between all @amsterdam repositories. Which should be predictable and equal on all repos.

Unnecessary logging done by `authorization_django` can be easily silenced with `settings.LOGGING` configuration and any additional logging can be added in our own middlewares.

My suggestions is to revert change related to `is_authorized_for`, as improvements are not as significant, compared to possible future issues.

PR #175, opened 2021-07-20
Fix field ordering in amsterdam schema

vdboor:
This also fixes the field ordering in the API. JSONField is converted back to TextField
Needs an `import_schemas` to update the contents of the database after deployment.


src/schematools/types.py
```diff
@@ -486,9 +486,9 @@ def get_fields(self, include_sub_fields: bool = False) -> Iterator[DatasetFieldS
         if self.has_compound_key:
             yield DatasetFieldSchema(_parent_table=self, _required=True, type="string", id="id")
 
-    @property
-    def fields(self) -> Iterator[DatasetFieldSchema]:
-        yield from self.get_fields(include_sub_fields=True)
+    @cached_property
```
jjmurre, line 489: Makes sense!

jjmurre (approved):
LGTM. Good to know that Jsonb does not retain ordering.

PR #177, opened 2021-07-21
AB#17588 Datasets with table refs load from file

Yrob:
Bugfix Datasets with tables in separate files (table references)
load correctly when read from a local file.

Dit zorgt dat `schema validate` op lokale datasets met losse tabellen werkt.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -6,6 +6,7 @@ from collections import UserDict
 from dataclasses import dataclass, field
 from enum import Enum
 from functools import cached_property, total_ordering
+from pathlib import Path
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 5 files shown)


src/schematools/types.py
```diff
@@ -94,7 +95,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
     def from_file(cls, filename: str) -> DatasetSchema:
         """Open an Amsterdam schema from a file."""
         with open(filename) as fh:
-            return cls.from_dict(json.load(fh))
+            dict = json.load(fh)
```
lbam, line 98: Using `dict` as a variable name is kind of confusing. `d`?

lbam (approved):
Aside from nitpick, looks fine.

src/schematools/types.py
```diff
@@ -94,7 +95,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
     def from_file(cls, filename: str) -> DatasetSchema:
         """Open an Amsterdam schema from a file."""
```
lbam, line 96: Maybe change the docstring too, because it can load multiple files?

PR #178, opened 2021-07-22
Loose relation and through id fix

jjmurre:
Two small schematools fixes. Explanation in de commit messages.

### DIFF
--- a/src/schematools/contrib/django/factories.py
+++ b/src/schematools/contrib/django/factories.py
@@ -99,7 +99,14 @@ class RelationMaker:
             (identifier_field.name, identifier_field.type),
         }
 
-        return destination_type_set > source_type_set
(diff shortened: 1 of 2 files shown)


src/schematools/contrib/django/factories.py
```diff
@@ -99,7 +99,7 @@ def is_loose_relation(cls, relation, dataset, field):
             (identifier_field.name, identifier_field.type),
         }
 
-        return source_type_set != destination_type_set
+        return not destination_type_set <= source_type_set
```
lbam, line 102:
```suggestion
        return destination_type_set >= source_type_set
```

jjmurre: Tx! Clearer.

PR #188, opened 2021-09-16
Fix foreigen keys to tables with camelCase (bbga.indicatorenDefinities)

vdboor:


vdboor:
was caused by my local branch

PR #192, opened 2021-10-06
Make sure AdditionalRelationSchema has the same relational properties

vdboor:
This allows both AdditionalRelationSchema and DatasetFieldSchema based relations to be accessed in the same way.


lbam (approved):

jjmurre (approved):
LGTM

PR #201, opened 2021-11-01
Validate change_dataset's --endpoint-url argument

lbam:
The argument needs to contain `{table_id}` for DSO-API to insert the actual table name.

### DIFF
--- a/src/schematools/contrib/django/management/commands/change_dataset.py
+++ b/src/schematools/contrib/django/management/commands/change_dataset.py
@@ -1,5 +1,7 @@
+import argparse
 from argparse import ArgumentTypeError
 from distutils.util import strtobool
+from typing import Any
 
 from django.core.exceptions import ValidationError
 from django.core.management import BaseCommand, CommandError
@@ -7,15 +9,15 @@ from django.core.management import BaseCommand, CommandError
 from schematools.contrib.django.models import Dataset
 
 
-def _strtobool(value):
+def _strtobool(value: Any) -> bool:
     try:
         return bool(strtobool(value))
     except ValueError:
         raise ArgumentTypeError("expected boolean value") from None
 
 
-class Command(BaseCommand):
-    help = "Modify the settings for a dataset."
+class Command(BaseCommand):  # noqa: D101
+    help = "Modify the settings for a dataset."  # noqa: A003
     requires_system_checks = False
     setting_options = (
         "auth",
@@ -26,7 +28,7 @@ class Command(BaseCommand):
         "path",
     )
 
-    def add_arguments(self, parser):
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:  # noqa: D102
         parser.add_argument("dataset", help="Name of the dataset")
         parser.add_argument(
             "--enable-db",
@@ -60,7 +62,7 @@ class Command(BaseCommand):
         parser.add_argument("--ordering", type=int, help="Set the ordering of the dataset")
         parser.add_argument("--endpoint-url")
 
-    def handle(self, *args, **options):
+    def handle(self, *args: Any, **options: Any) -> None:  # noqa: D102
         name = options.pop("dataset")
         try:
             dataset = Dataset.objects.get(name=name)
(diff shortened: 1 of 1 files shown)


jjmurre (approved):
LGTM. Good idea to add this extra precaution.

src/schematools/contrib/django/management/commands/change_dataset.py
```diff
@@ -1,21 +1,23 @@
...
+from typing import Any
 
 from django.core.exceptions import ValidationError
 from django.core.management import BaseCommand, CommandError
 
 from schematools.contrib.django.models import Dataset
 
 
-def _strtobool(value):
+def _strtobool(value: Any) -> bool:
```
gkoller, line 12: Judging by its name `strtobool` this function converts a `str` to a `bool`. The docstring of the function you delegate to confirm that: _Convert a string representation of truth to true (1) or false (0)._. Hence the type annotation of `value` should not be `Any`, but `str`

src/schematools/contrib/django/management/commands/change_dataset.py
```diff
@@ -1,21 +1,23 @@
...
     try:
         return bool(strtobool(value))
     except ValueError:
         raise ArgumentTypeError("expected boolean value") from None
```
gkoller, line 16: I realize this is existing code, but the message is incorrect. A boolean value was NOT expected. On the contrary the boolean value is what this function is supposed to be generating. An arguably better message would be: _Expected a truth value_. Which is basically the message of the `ValueError` exception. Hence my suggestion would be to change this into:

```python
except ValueError as ve:
            raise ArgumentTypeError(str(ve)) from None
```

src/schematools/contrib/django/management/commands/change_dataset.py
```diff
@@ -69,13 +71,19 @@ def handle(self, *args, **options):
             raise CommandError(f"Dataset not found: {name}.\nAvailable are: {available}") from None
 
         # Validate illogical combinations
-        if options.get("endpoint_url"):
+        endpoint_url = options.get("endpoint_url")
+        if endpoint_url is not None:
```
gkoller, line 75: More idiomatic:

```python
if endpoint_url:
    ..
```.

File: src/schematools/contrib/django/management/commands/change_dataset.py [hunk 4]
```diff
@@ -69,13 +71,19 @@ def handle(self, *args, **options):
...
             if options.get("enable_db"):
                 raise CommandError("Can't use --endpoint-url with --enable-db")
 
             if dataset.enable_db:
                 options["enable_db"] = False
 
+        # URL endpoints need to contain {table_id}, which DSO-API replaces
+        # with the actual table name.
+        if endpoint_url is not None and "{table_id}" not in endpoint_url:
```
gkoller, line 84: More idiomatic:

```python
if endpoint_url and "{table_id}" not in endpoint_url:
    ...
```

src/schematools/contrib/django/management/commands/change_dataset.py
```diff
@@ -69,13 +71,19 @@ def handle(self, *args, **options):
...
 
         # Validate illogical combinations
-        if options.get("endpoint_url"):
+        endpoint_url = options.get("endpoint_url")
+        if endpoint_url is not None:
             if options.get("enable_db"):
                 raise CommandError("Can't use --endpoint-url with --enable-db")
 
             if dataset.enable_db:
                 options["enable_db"] = False
```
gkoller, line 80: No need for an if-statement:

```python
options["enable_db"] = dataset.enable_db
```

gkoller: Existing code, I know, but no harm in cleaning up code when we (almost) hit it.

src/schematools/contrib/django/management/commands/change_dataset.py
lbam: Fixed in 94e532695932246f7fa2a57e090f18b01043a59d.

PR #202, opened 2021-11-02
Improve typing annotations for DatasetTableSchema.temporal

vdboor:
Using a NamedTuple instead of the `Tuple[str, str]` logic.


gkoller (approved):
Nice. Much more descriptive.

PR #203, opened 2021-11-02
AB#20011 Change `field.sub_fields` to not add prefixes

jjmurre:
The implementation of `field.sub_fields` was adding prefixes
to fieldnames. These prefixes are for a special use-case,
where the `sub_fields` method is used to generate extra fields
describing the components of a compound foreign key.

However, adding these prefixes should not be the default behaviour of `sub_fields`.

This has been fixed. The property fields `sub_fields`
has now been converted into a function `get_sub_fields`
with an extra parameter `add_prefixes`.

The property `sub_fields` is still available and implements
the default use-case, generating fields without prefixes.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -522,6 +522,13 @@ class DatasetTableSchema(SchemaType):
         return self.get("description")
 
     def get_fields(self, include_sub_fields: bool = False) -> Iterator[DatasetFieldSchema]:
+        """Get the fields for this table.
+
+        Args:
+            include_sub_fields: Merge the sub fields of an FK relation into the fields
+            of this table. The ids of these fields need to be prefixed
+            (usually with the `id` of the relation field) to avoid name collisions.
+        """
         required = set(self["schema"]["required"])
         for id_, spec in self["schema"]["properties"].items():
             field_schema = DatasetFieldSchema(
@@ -534,8 +541,11 @@ class DatasetTableSchema(SchemaType):
             # These fields are added to identify the different
             # components of a compound FK to a another table
             if field_schema.relation is not None and field_schema.is_object and include_sub_fields:
-                for subfield_schema in field_schema.sub_fields:
-                    yield subfield_schema
+                for sub_field in field_schema.get_sub_fields(add_prefixes=True):
+                    # We exclude temporal fields, they need not to be merged into the table fields
+                    if sub_field.is_temporal:
+                        continue
+                    yield sub_field
             yield field_schema
 
         # If compound key, add PK field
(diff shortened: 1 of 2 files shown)


vdboor (approved):
This should fix the docs, nicely done.
I do think this can be refined, as `table.fields` will still return all fields with a prefix. It would be better if the caller would call `field.db_name` (or generate that themselves) if they want the prefixed version.

src/schematools/types.py
```diff
@@ -533,7 +533,7 @@ def get_fields(self, include_sub_fields: bool = False) -> Iterator[DatasetFieldS
             # These fields are added to identify the different
             # components of a compound FK to a another table
             if field_schema.relation is not None and field_schema.is_object and include_sub_fields:
-                for subfield_schema in field_schema.sub_fields:
+                for subfield_schema in field_schema.get_sub_fields(add_prefixes=True):
                     yield subfield_schema
```
vdboor, line 537:
```suggestion
                yield from field_schema.get_sub_fields(add_prefixes=True)
```

src/schematools/types.py
```diff
@@ -989,12 +989,24 @@ def get_dimension_fieldnames(self) -> Dict[str, Tuple[str, str]]:
     def sub_fields(self) -> Iterator[DatasetFieldSchema]:
         """Return the sub fields for a nested structure.
 
+        Calls the `get_sub_fields` method without argument,
+        so no prefixes are added to the field ids.
+        This is the default situation.
+        """
+        return self.get_sub_fields()
```
vdboor, line 996:
```suggestion
    def sub_fields(self) -> List[DatasetFieldSchema]:
        """Return the sub fields for a nested structure.

        Calls the `get_sub_fields` method without argument,
        so no prefixes are added to the field ids.
        This is the default situation.
        """
        return list(self.get_sub_fields())
```

and make it a `@cached_property`

jjmurre: Good idea, gets called quite often.

jjmurre:
> This should fix the docs, nicely done. I do think this can be refined, as `table.fields` will still return all fields with a prefix. It would be better if the caller would call `field.db_name` (or generate that themselves) if they want the prefixed version.

At the moment, that would break things. Because the extra fields that are added when `include_sub_fields` is `True` are merged into the list of fields of the tables. So, this would lead to name clashes (the `identificatie`, `volgnummer` of the main table could not be discriminated from the `identificatie` and `volgnummer` of the relation that is being merge into the list of fields.

May the `included_sub_fields` argments when calling `table.get_fields` from `table.fields` should change to `False`.
If done now, a lot of things are breaking. So, a change should be done, where at a number of places a call to `table.fields` should change to `table.get_fields(include_sub_fields=True)`. I'll leave that to a separate refactoring.

PR #206, opened 2021-11-04
Decouple LooseRelationField from schema's. Pass a to_field instead.

vdboor:
Sometimes I found during the fixing of temporality-listings. The loose relations M2M field doesn't really do anything loose at all. I guess the relations are created at import?

The constructor receives a 'relation' argument and 'max_length' but these are not used or don't do anything different from standard Django. The same goes for `LooseRelationManyToManyField`, which didn't do anything with the provided max_length. And the 'related_model' was not used at all either.


jjmurre (approved):
Good cleanup. There were some left-overs from the initial implementation.

PR #207, opened 2021-11-08
AB#23066 Add configurable loader for datasets

jjmurre:
See the first commit message for explanation about this PR.

### DIFF
--- a/src/schematools/datasetcollection.py
+++ b/src/schematools/datasetcollection.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import os
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING, Dict, Optional, Union, cast
 from urllib.parse import urlparse
 
 from more_ds.network.url import URL
@@ -30,14 +30,14 @@ class DatasetCollection(metaclass=Singleton):
             schema_loader: An alternative schema loader can be provided
                 If schema_loader is None, the default url loader is used.
         """
-        self.datasets_cache = {}
-        self.schema_loader = None
+        self.datasets_cache: Dict[str, DatasetSchema] = {}
+        self.schema_loader: Optional[loaders.SchemaLoader] = None
 
-    def set_schema_loader(self, schema_loader: loaders.SchemaLoader):
+    def set_schema_loader(self, schema_loader: loaders.SchemaLoader) -> None:
         """Set the schema loader for the datasetcollection."""
         self.schema_loader = schema_loader
 
-    def get_schema_loader(self):
+    def get_schema_loader(self) -> loaders.SchemaLoader:
         """Get the schema_loader."""
         if self.schema_loader is None:
             raise ValueError("The datasetcollection should be initialized with a schema loader")
@@ -71,7 +71,7 @@ class DatasetCollection(metaclass=Singleton):
             return dataset
 
 
-def set_schema_loader(schema_url: Union[URL, str]):
+def set_schema_loader(schema_url: Union[URL, str]) -> None:
     """Initialize the schema loader at module load time.
 
     schema_url:
@@ -79,6 +79,7 @@ def set_schema_loader(schema_url: Union[URL, str]):
         can be a web url, or a filesystem path.
     """
     dataset_collection = DatasetCollection()
+    loader: Optional[loaders.SchemaLoader] = None  # pleasing mypy
     if urlparse(schema_url).scheme in ("http", "https"):
         loader = loaders.URLSchemaLoader(schema_url)
     else:
--- a/src/schematools/datasetcollection.py
+++ b/src/schematools/datasetcollection.py
@@ -1,12 +1,18 @@
 from __future__ import annotations
 
 import os
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Union
+from urllib.parse import urlparse
 
+from more_ds.network.url import URL
 from simple_singleton import Singleton
 
 if TYPE_CHECKING:
     from schematools.types import DatasetSchema
+
+from schematools import DEFAULT_SCHEMA_URL, loaders
+
+# Initialize the schema_loader in the DatasetCollection singleton
 
 
 class DatasetCollection(metaclass=Singleton):
@@ -18,27 +24,36 @@ class DatasetCollection(metaclass=Singleton):
     """
 
     def __init__(self) -> None:
-        self.datasets_cache = {}
+        """Initialize the DatasetCollection.
 
-    def _load_dataset(self, dataset_id: str) -> Optional[DatasetSchema]:
-        """Loads the dataset from SCHEMA_URL.
-
-        If SCHEMA_URL is not defined, return None.
+        Args:
+            schema_loader: An alternative schema loader can be provided
+                If schema_loader is None, the default url loader is used.
         """
-        # Avoid circular import problem.
-        from schematools.utils import dataset_schema_from_url
+        self.datasets_cache = {}
+        self.schema_loader = None
 
-        try:
-            schemas_url = os.environ["SCHEMA_URL"]
-        except KeyError:
+    def set_schema_loader(self, schema_loader: loaders.SchemaLoader):
+        """Set the schema loader for the datasetcollection."""
+        self.schema_loader = schema_loader
+
+    def get_schema_loader(self):
+        """Get the schema_loader."""
+        if self.schema_loader is None:
+            raise ValueError("The datasetcollection should be initialized with a schema loader")
+        return self.schema_loader
+
+    def _load_dataset(self, dataset_id: str, prefetch_related: bool) -> Optional[DatasetSchema]:
+        """Loads the dataset, using the configured loader."""
+        if self.schema_loader is None:
             return None
-
-        return dataset_schema_from_url(schemas_url, dataset_id, prefetch_related=True)
+        return self.schema_loader.get_dataset(dataset_id, prefetch_related=prefetch_related)
 
     def add_dataset(self, dataset: DatasetSchema) -> None:
+        """Add a dataset to the cache."""
         self.datasets_cache[dataset.id] = dataset
 
-    def get_dataset(self, dataset_id: str) -> DatasetSchema:
+    def get_dataset(self, dataset_id: str, prefetch_related: bool = False) -> DatasetSchema:
         """Gets a dataset by id from the cache.
 
         If not available, load the dataset from the SCHEMA_URL location.
@@ -49,8 +64,30 @@ class DatasetCollection(metaclass=Singleton):
         try:
             return self.datasets_cache[dataset_id]
         except KeyError:
-            dataset = self._load_dataset(dataset_id)
+            dataset = self._load_dataset(dataset_id, prefetch_related=prefetch_related)
             if dataset is None:
                 raise ValueError(f"Dataset {dataset_id} is missing.") from None
             self.add_dataset(dataset)
             return dataset
+
+
+def set_schema_loader(schema_url: Union[URL, str]):
+    """Initialize the schema loader at module load time.
+
+    schema_url:
+        Location where the schemas can be found. This
+        can be a web url, or a filesystem path.
+    """
+    dataset_collection = DatasetCollection()
+    if urlparse(schema_url).scheme in ("http", "https"):
+        loader = loaders.URLSchemaLoader(schema_url)
+    else:
+        loader = loaders.FileSystemSchemaLoader(schema_url)
+    dataset_collection.set_schema_loader(loader)
+
+
+# The scheme loader is initialized from the `SCHEMA_URL` environment variable,
+# or from the DEFAULT_SCHEMA_URL constant.
+# This call is done at module load time, to have an initial value for the schemaloader.
+# If needed, an alternative schemaloader can be injected into `DatasetCollection` at runtime.
+set_schema_loader(os.environ.get("SCHEMA_URL", DEFAULT_SCHEMA_URL))
--- a/src/schematools/loaders.py
+++ b/src/schematools/loaders.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Union
+
+from more_ds.network.url import URL
+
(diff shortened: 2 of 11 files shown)


lbam:
I don't really like the way we're hiding the configuration of library code in the environment. Even if we want to do that, we should be explicit about it.

src/schematools/loaders.py
```diff
@@ -0,0 +1,35 @@
...
+
+def get_loader(reference: str) -> Callable:
+    """Gets a loader callable from a string reference."""
+    module_name, separator, callable_name = reference.partition(":")
+    if not separator:
+        raise ValueError("Provide a fully qualified reference `package.module:callable`")
+    obj = importlib.import_module(module_name)
+    for attr in callable_name.split("."):
+        obj = getattr(obj, attr)
+    return obj
```
lbam, line 17: This looks like a security breach waiting to happen.

jjmurre:
> I don't really like the way we're hiding the configuration of library code in the environment. If we do want to do that, we should be explicit about it. The new docstrings are pretty vague about what is going on and never mention either `SCHEMA_LOADER` or `SCHEMA_PATH`.

Would it just be a question of better explanation in the docstrings? We need to uses environment vars to configure the `github` actions. But, we could make an explicit mapping for discrete loaders. I agree that this would be safer.

src/schematools/loaders.py
jjmurre: I think you're right, because this is used in the context of github actions. We do not know enough about that environment to be sure that no one touches the environment vars.

lbam:
Can we not switch on the scheme part of `SCHEMA_URL` (`https` or `file`) and determine what do to based on it? At least that way, we have only one environment variable to document instead of three interacting ones.

jjmurre:
> Can we not switch on the scheme part of `SCHEMA_URL` (`https` or `file`) and determine what do to based on it? At least that way, we have only one environment variable to document instead of three interacting ones.

I thought of that. However `_URL` can be a bit misleading. However, when using `file:/// ...` it would be logical. Only thing is that a `file:// ..` url cannot be relative afaik. And that is what we need in the github action.

src/schematools/loaders.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def schema_file_loader(dataset_id: str) -> Optional[DatasetSchema]:
+    """Load a datasetschema from the filesystem."""
+    try:
+        schemas_path = os.environ["SCHEMA_PATH"]
```
gkoller, line 23: Could we consider using Pydantic for dealing with settings and environment variables: https://pydantic-docs.helpmanual.io/usage/settings/

lbam:
`file:` URLs can be anything we want, because we interpret them :) They're awkward to use, though.

Even easier is to just treat the "URL" as a filename if it doesn't start with `http` or `https`:

```python
u = urlparse(location)
if u.scheme in ("http", "https"):
    # treat location as URL
else:
    # treat location as local path
```

src/schematools/loaders.py
```diff
@@ -0,0 +1,35 @@
...
+    return dataset_schema_from_file(dataset_id, schemas_path=schemas_path, prefetch_related=True)
+
+
+def schema_url_loader(dataset_id: str) -> Optional[DatasetSchema]:
+    """Load a datasetschema from the a url."""
+    try:
+        schemas_url = os.environ["SCHEMA_URL"]
+    except KeyError:
+        return None
+    return dataset_schema_from_url(schemas_url, dataset_id, prefetch_related=True)
```
gkoller, line 35: There is a discrepancy between loading schema's from URL versus files:
* `dataset_schema_from_url` calls `schematools.utils.schema_from_url`: a utility function.
* `dataset_schema_from_file` calls `DatasetSchema.from_file`: a member function defined on the `DatasetSchema` class

Why is one a member function and the other simply a utility function? Can we remedy this discrepancy in this PR?

I would argue that loading a schema, either from file or URL should not be defined in a member function on `DatasetSchema`. Hence I thing `DatasetSchema.from_file` should become an utility function as well.

src/schematools/loaders.py
jjmurre: > `file:` URLs can be anything we want, because we interpret them :) They're awkward to use, though.
> 
> Even easier is to just treat the "URL" as a filename if it doesn't start with `http` or `https`:
> 
> ```python
> u = urlparse(location)
> if u.scheme in ("http", "https"):
>     # treat location as URL
> else:
>     # treat location as local path
> ```
> 
> (or just do `location.startswith("https://") or location.startswith("http://")`)

Makes it simpler and is easy to understand. And, that's what is also being done in the cli code.

jjmurre: +1 for Pydantic. But, this should be used for all enviroment acces in the whole repo. Agreed?

src/schematools/loaders.py
jjmurre: I'll look into that. This would be a breaking change, so this would cause a major version bump.

jjmurre: Just a question, what's against a `DatasetSchema.from_file`? This class-method-as-factory is quite common practice in python code, or am I wrong?

src/schematools/datasetcollection.py
```diff
@@ -21,19 +21,16 @@ def __init__(self) -> None:
...
+        # Late import, avoid circular dependency.
+        from schematools.loaders import get_loader, schema_file_loader, schema_url_loader
 
-        try:
-            schemas_url = os.environ["SCHEMA_URL"]
-        except KeyError:
-            return None
-
-        return dataset_schema_from_url(schemas_url, dataset_id, prefetch_related=True)
+        schema_loader = os.environ.get("SCHEMA_LOADER")
```
vdboor, line 31: waarom ben je hier een `SCHEMA_LOADER` aan het resolven?
Als je datasets ophaalt, begin je met een loader. Die maakt een `DatasetCollection` aan en de bijgehorende dataset. Vervolgens kan de collection `self.loader.get_dataset(...)` uitvoeren.

src/schematools/datasetcollection.py
```diff
@@ -21,19 +21,16 @@ def __init__(self) -> None:
...
+        loader = schema_url_loader if schema_loader is None else get_loader(schema_loader)
+        return loader(dataset_id)
```
vdboor, line 33: ik zou van de loader geen callable maken, maar een class. Zie Django's template loaders voor een voorbeeld. Zo'n class biedt iets meer mogelijkheden tot state bijhouden, beter annotatiemogelijkheden en uitbreidbaarheid.

src/schematools/loaders.py
lbam: The data structure doesn't know about files, just the factory. Where you put the factory only determines how you call it, and how easy it is to discover.

gkoller: > Where you put the factory only determines how you call it

I tend to disagree. Classes are a way to encapsulate data with behavior on that data. A class representing an Amsterdam Schema dataset does not need any logic for obtaining the source of that data from the filesystem, from the network, from a bluetooth connection, or from whatever data store that comes to mind.

That kind of logic - logic for dealing with external systems - has absolutely nothing to do with the dataset data structure. As such it does not belong in the dataset class.

But as I said to @jjmurre I care for consistency more than in this specific case. Either we put the logic for loading from file system and loading from the network both on the class or  both as utility functions outside of the class. Not one on the class and the other as a utility function.

gkoller: Basically what I am saying is that methods related to I/O don't belong on the class `DatasetSchema` as `DatasetSchema` is not about I/O in any way

src/schematools/loaders.py
gkoller: Agreed

src/schematools/loaders.py
```diff
@@ -0,0 +1,54 @@
...
+
+from more_ds.network.url import URL
+
+if TYPE_CHECKING:
+    from schematools.types import DatasetSchema
+
+
+class SchemaLoader:
+    """Base class for schema loaders."""
+
```
gkoller, line 14: Define a instance variable `schema_url: Union[Path, URL]` according to https://www.python.org/dev/peps/pep-0526/

src/schematools/loaders.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, schema_url: Union[Path, URL, str]):
+        """Initialize the schema loader.
+
+        schema_url:
+            Either a web url to the dataset schemas, or a path on the local filesystem.
+        """
+        self.schema_url = schema_url
```
gkoller, line 21:
```python
self.schema_url = schema_url if isinstance(schema_url, Path) else URL(schema_url)
```

Daarna is het altijd een `Path` of een `URL` (en `URL` kan overal gebruikt worden waar een `str` wordt gebruikt.)

src/schematools/datasetcollection.py
```diff
@@ -18,27 +24,35 @@ class DatasetCollection(metaclass=Singleton):
...
 
-        If SCHEMA_URL is not defined, return None.
+        Args:
+            schema_loader: An alternative schema loader can be provided
+                If schema_loader is None, the default url loader is used.
         """
-        # Avoid circular import problem.
-        from schematools.utils import dataset_schema_from_url
+        self.datasets_cache: Dict[str, DatasetSchema] = {}
+        self.schema_loader: Optional[loaders.SchemaLoader] = None
```
gkoller, line 34: Maybe these should be defined as instance variables according to https://www.python.org/dev/peps/pep-0526/

src/schematools/datasetcollection.py
```diff
@@ -18,27 +24,35 @@ class DatasetCollection(metaclass=Singleton):
...
+    def get_schema_loader(self) -> loaders.SchemaLoader:
+        """Get the schema_loader."""
+        if self.schema_loader is None:
+            raise ValueError("The datasetcollection should be initialized with a schema loader")
+        return self.schema_loader
 
-        return dataset_schema_from_url(schemas_url, dataset_id, prefetch_related=True)
+    def _load_dataset(self, dataset_id: str, prefetch_related: bool) -> Optional[DatasetSchema]:
+        """Loads the dataset, using the configured loader."""
+        schema_loader = cast(loaders.SchemaLoader, self.schema_loader)
```
gkoller, line 48: If you have `schema_loader` defined as an instance variable (see https://github.com/Amsterdam/schema-tools/pull/207/files#r749223188) You don't need the `cast` anymore. A simple `is not None` check would be sufficient in that case.

src/schematools/types.py
```diff
@@ -94,20 +98,16 @@ def __repr__(self):
         return f"<{self.__class__.__name__}: {self['id']}>"
 
     @classmethod
+    @deprecated(
```
gkoller, line 101: Nice!! :+1:

gkoller (approved):
Clean code. Like it :+1:

src/schematools/loaders.py
jjmurre: Good point, tx!

src/schematools/types.py
```diff
@@ -94,20 +98,16 @@ def __repr__(self):
...
+        version="2.3.1",
+        reason="""The `DatasetSchema.from_file` is replaced by
```
lbam, line 103:
```suggestion
        reason="""The `DatasetSchema.from_file` has been replaced by
```

src/schematools/datasetcollection.py
```diff
@@ -49,8 +63,29 @@ def get_dataset(self, dataset_id: str) -> DatasetSchema:
...
+    if urlparse(schema_url).scheme in ("http", "https"):
+        loader = loaders.URLSchemaLoader(schema_url)
+    else:
+        loader = loaders.FileSystemSchemaLoader(schema_url)
+    dataset_collection.set_schema_loader(loader)
+
+
+# The scheme loader is initialized from the `SCHEMA_URL` environment variable,
+# or from the DEFAULT_SCHEMA_URL.
+set_schema_loader(os.environ.get("SCHEMA_URL", DEFAULT_SCHEMA_URL))
```
lbam, line 91: If this is meant to run only at module init time, shouldn't it be more private?

```suggestion
def _set_schema_loader(schema_url: Union[URL, str]) -> None:
    """Initialize the schema loader at module load time.

    schema_url:
        Location where the schemas can be found. This
        can be a web url, or a filesystem path.
    """
    dataset_collection = DatasetCollection()
    loader: Optional[loaders.SchemaLoader] = None  # pleasing mypy
    if urlparse(schema_url).scheme in ("http", "https"):
        loader = loaders.URLSchemaLoader(schema_url)
    else:
        loader = loaders.FileSystemSchemaLoader(schema_url)
    dataset_collection.set_schema_loader(loader)


# The scheme loader is initialized from the `SCHEMA_URL` environment variable,
# or from the DEFAULT_SCHEMA_URL.
_set_schema_loader(os.environ.get("SCHEMA_URL", DEFAULT_SCHEMA_URL))
```

src/schematools/types.py
```diff
@@ -311,7 +311,6 @@ def build_through_table(
         from schematools.utils import get_rel_table_identifier, to_snake_case, toCamelCase
 
         def _get_fk_target_table(dataset_id: str, table_id: str) -> DatasetTableSchema:
-            """Gets the"""
```
lbam, line 314: I was once given a largeish codebase to maintain after its author had left the company. Apparently, they had turned some autocomment feature on in their IDE, and blindly committed everything it produced, so every method had an docstring like

```python
def get_table_from_file(table, filename, ignore_errors):
    """Gets the table.

    Args:
        table: The table.
        filename: The name.
        ignore_errors: The ignore errors.
    """
```

There was also a method called `get`, with this exact docstring :)

src/schematools/datasetcollection.py
lbam: Shouldn't these also be private, `_datasets_cache` etc.?

src/schematools/datasetcollection.py
jjmurre: Good point to express that is it is not meant to be called externally by flagging it private with `_`.

jjmurre: # This call is done at module load time, to have an initial value for the schemaloader.
# If needed, an alternative schemaloader can be injected into `DatasetCollection` at runtime.

PR #208, opened 2021-11-09
Make get_field_schema() easier to find

vdboor:
Inspired from feedback from Kees when working on the first tickets.

Moving get_field_schema() call to a place where it would be expected 
This should make it easier to find such logic for newcomers, as things
like get_table_schema() are also found in the same place.

### DIFF
--- a/src/schematools/contrib/django/models.py
+++ b/src/schematools/contrib/django/models.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 import json
 import logging
 import re
+import warnings
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 from django.apps import apps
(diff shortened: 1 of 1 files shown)


src/schematools/contrib/django/models.py
```diff
@@ -162,6 +163,32 @@ def table_schema(cls) -> DatasetTableSchema:
         """Give access to the original table_schema that this model implements."""
         return cls._table_schema
 
+    @classmethod
+    def get_field_schema(cls, model_field: Union[models.Field, models.ForeignObjectRel]):
+        """Provide access to the underlying amsterdam schema field that created the model field."""
+        if isinstance(model_field, models.ForeignObjectRel):
+            # When Django auto-creates the related field, it doesn't have `field_schema`,
+            # but it can be resolved by looking up the original forward relation.
+            return model_field.remote_field.field_schema
```
DonQueso89, line 172: is it guaranteed that the forward relation has the `field_schema` property? or do we need the same exception handler as in the else-branch ?

DonQueso89 (approved):

lbam (approved):

src/schematools/contrib/django/models.py
vdboor: Good point, I had my doubts too. Updated the PR.

PR #209, opened 2021-11-11
explicit database schema default when None

cvriel:
The method tables_factory retrieves database schema if defined as a parameter when creating a database table. If schema is None, it now defaults to Public. Leave it to None will have a risk that the DB schema that is currently in use will be used to create the table in leading to unwanted/unexepected results.

### DIFF
--- a/src/schematools/__init__.py
+++ b/src/schematools/__init__.py
@@ -4,4 +4,4 @@ RELATION_INDICATOR: Final[str] = "_"
 MAX_TABLE_NAME_LENGTH: Final[int] = 63  # limitation of PostgreSQL
 TMP_TABLE_POSTFIX: Final[str] = "_new"
 TABLE_INDEX_POSTFIX: Final[str] = "_idx"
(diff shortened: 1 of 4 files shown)


src/schematools/__init__.py
```diff
@@ -4,3 +4,4 @@
 MAX_TABLE_NAME_LENGTH: Final[int] = 63  # limitation of PostgreSQL
 TMP_TABLE_POSTFIX: Final[str] = "_new"
 TABLE_INDEX_POSTFIX: Final[str] = "_idx"
+DATABASE_SCHEMA_NAME_DEFAULT: Final[str] = "PUBLIC"
```
jjmurre, line 7: Just a question, why schema names in all caps? Is'nt is more common practice to lowercase all identifiers for PostgreSQL?

jjmurre (approved):
Good idea to be more explicit!

src/schematools/__init__.py
cvriel: You are right, no reason to make it upper, lower cased it

PR #211, opened 2021-11-15
Replace DatasetFieldSchema.sub_fields by subfields

lbam:
This is the proper spelling.

More importantly, I wanted to change the semantics of the method and property, and renaming it seemed the easiest option to break any code that relied on the old behavior and force reconsideration.

The new method and property return an empty iterator/list for an array or object instead of raising a `ValueError`. A use site in DSO-API caught the `ValueError`, but as a result also caught unrelated errors, causing a bug in production and tests successes where failures were appropriate (Amsterdam/dso-api#402). It looks like the only other call site performs a custom type check and should not be affected.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -447,7 +447,7 @@ class DatasetSchema(SchemaType):
 
         related_ids = []
         for table in self.tables:
-            for field in table.get_fields(include_sub_fields=False):
+            for field in table.get_fields(include_subfields=False):
                 a_relation = field.relation or field.nm_relation
                 if a_relation is not None:
                     dataset_id, table_id = a_relation.split(":")
@@ -523,11 +523,11 @@ class DatasetTableSchema(SchemaType):
         """The description of the table as stated in the schema."""
         return self.get("description")
 
-    def get_fields(self, include_sub_fields: bool = False) -> Iterator[DatasetFieldSchema]:
+    def get_fields(self, include_subfields: bool = False) -> Iterator[DatasetFieldSchema]:
         """Get the fields for this table.
 
         Args:
-            include_sub_fields: Merge the sub fields of an FK relation into the fields
+            include_subfields: Merge the subfields of an FK relation into the fields
             of this table. The ids of these fields need to be prefixed
             (usually with the `id` of the relation field) to avoid name collisions.
         """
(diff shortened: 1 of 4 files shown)


src/schematools/types.py
```diff
@@ -1018,21 +1018,23 @@ def get_sub_fields(self, add_prefixes=False) -> Iterator[DatasetFieldSchema]:
...
+            # Field has direct subfields (type=object)
             required = set(self.get("required", []))
             properties = self["properties"]
         elif self.is_array_of_objects and self.field_items is not None:
             # Field has an array of objects (type=array, items are objects)
             required = set(self.field_items.get("required") or ())
             properties = self.field_items["properties"]
         else:
-            raise ValueError("Subfields are only possible for 'object' or 'array' fields.")
+            return ()
```
gkoller, line 1037: This returns an `Iterable` not an `Iterator`. You should adjust the return type of this method.

src/schematools/types.py
```diff
@@ -995,17 +995,17 @@ def get_dimension_fieldnames(self) -> Dict[str, TemporalDimensionFields]:
...
+        Calls the `get_subfields` method without argument,
         so no prefixes are added to the field ids.
         This is the default situation.
         """
-        return list(self.get_sub_fields())
+        return list(self.get_subfields())
 
-    def get_sub_fields(self, add_prefixes=False) -> Iterator[DatasetFieldSchema]:
-        """Return the sub fields for a nested structure.
+    def get_subfields(self, add_prefixes=False) -> Iterator[DatasetFieldSchema]:
```
gkoller, line 1007: See https://github.com/Amsterdam/schema-tools/pull/211/files#r749252894

src/schematools/types.py
jjmurre: Hm. I am a bit hesistant about this one. When a field is not a relation- or nested field, the caller is not supposed to call `get_sub_fields` on that field, hence the `ValueError`. Returning an empty tuple silences this error. Maybe that is not a problem, but the caller is not informed that she is doing something weird. Could this not become problematic?

lbam: With this behavior, both call sites in dso-api work without any checks or try-catch constructs. Any behavior we choose can cause problems, but empirically, this is safer than the current behavior (N=1, but still :)

lbam: @gkoller Changed to `Iterable`.

jjmurre: Ok, fine with me. Maybe good to  add a warning in the log.

PR #212, opened 2021-11-15
Only return surrogate pk for schemas with compound pk if schema does not define `id`

DonQueso89:
`get_fields()` now returns `id` twice for for example the verblijfsobjecten dataset

### DIFF
--- a/tests/files/compound_key.json
+++ b/tests/files/compound_key.json
@@ -17,7 +17,7 @@
         }
       },
       "schema": {
-        "$id": "https://github.com/Amsterdam/schemas/afracnswallows",
+        "$id": "https://github.com/Amsterdam/schemas/africanswallows",
         "$schema": "http://json-schema.org/draft-07/schema#",
         "type": "object",
         "additionalProperties": false,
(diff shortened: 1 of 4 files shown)


lbam:
Test?

lbam:
Nice, but I think you forgot to check in the `compound_key.json`?

DonQueso89:
> Nice, but I think you forgot to check in the `compound_key.json`?

whops

tests/files/compound_key.json
```diff
@@ -0,0 +1,55 @@
...
+      "type": "table",
+      "version": "0.0.1",
+      "temporal": {
+        "identifier": "volgnummer",
+        "dimensions": {
+          "geldigOp": ["beginGeldigheid", "eindGeldigheid"]
+        }
+      },
+      "schema": {
+        "$id": "https://github.com/Amsterdam/schemas/afracnswallows",
```
lbam, line 20:
```suggestion
        "$id": "https://github.com/Amsterdam/schemas/africanswallows",
```

lbam (approved):

PR #213, opened 2021-11-17
AB#24402 Deprecate methods dealing with additionalFilters

lbam:
Additional filters have been replaced by temporal ones.

I wanted to remove these in 3.0, but didn't have the required DSO-API changes ready. For 4.0, then.


jjmurre (approved):
LGTM. You could add a version to the deprecated decorator, so the user knows from which version the deprecation started.

PR #214, opened 2021-11-17
flake8 and mypy cleanup in types.py

lbam:
`flake8` on this file now passes, though the pre-commit hook will still find a few errors.

Number of `mypy` errors down from 85 to 53.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -111,7 +111,7 @@ class DatasetSchema(SchemaType):
         return dataset_schema_from_path(filename)
 
     @classmethod
-    def from_dict(cls, obj: Json) -> DatasetSchema:
+    def from_dict(cls, obj: Dict[str, Any]) -> DatasetSchema:
         """Parses given dict and validates the given schema"""
         if obj.get("type") != "dataset" or not isinstance(obj.get("tables"), list):
             raise ValueError("Invalid Amsterdam Dataset schema file")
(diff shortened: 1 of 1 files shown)


src/schematools/types.py
```diff
@@ -327,6 +327,7 @@ def _expand_relation_spec(
             expanded into an object.
             """
             if fk_target_table:
+                sub_table_schema = cast(sub_table_schema, Dict[str, Any])
```
gkoller, line 330: You should swap the arguments.

src/schematools/types.py
```diff
@@ -351,13 +352,10 @@ def _expand_relation_spec(
...
         ]
 
-        # XXX maybe not logical to snakecase the fieldname here.
-        # this is still schema-land.
-        snakecased_fieldname = to_snake_case(field.name)
         snakecased_field_id = to_snake_case(field.id)
         table_id = get_rel_table_identifier(len(self.id) + 1, table.id, snakecased_field_id)
 
-        sub_table_schema: Json = {
+        sub_table_schema: Dict[str, Any] = {
```
gkoller, line 358: Why this change. Surely the alias `Json` communicates the type of dictionary better then merely `Dict[str, Any]`?

gkoller (approved):
👍 for addressing these MyPy and flake8 issues!

src/schematools/types.py
lbam: https://github.com/Amsterdam/schema-tools/blob/75ad36b29d3f0b5c3e97d86c9233ad97546f5440/src/schematools/types.py#L39

Of these, only `Dict[str, Any]` supports the indexing by string below.

PR #217, opened 2021-11-22
Add cli command to generate SQL Create statements

jjmurre:
Some "external" users are starting to use `schematools`
to generate SQL tables from amsterdam schemas definitions.

However, the way the cli works at the moment is mainly
tailored to our use case and needs a full checkout of all the schemas.

This new cli command only needs an isolated amsterdam schema file,
and generates sql create commands from it.


gkoller (approved):
Nice addition 👍

PR #218, opened 2021-11-22
Guido/ab23973 validate active versions

gkoller:
Validation for `activeVersions` and table `id`s. See example below

Before we start messing with versioned table creation we should ensure that the datasets are consistent. Eg that:
- the active versions mentioned in a data set match the versions of the referenced tables.
- that the table `id`s in the dataset matches with the `id`s in the referenced tables.

A rather annoying problem I had to work around is that the the current spec for `activeVersions` does not allow for a dataset where all active version of a table are represented inline with the dataset. Eg, if a table has only one version it can be represented inline with the dataset: `dataset["tables"][0] = <table dict representation>`. However if there are multiple versions `dataset["tables"][0]` suddenly becomes an entirely different entity, namely a dict that looks like:

```python
 {
      "id": "someTable",
      "$ref": "someTable/v1.1.0",
      "activeVersions": {
        "1.1.0": "someTablel/v1.1.0"
        "1.0.0": "someTablel/v1.0.0"
      }
```
This lack of uniformity complicates things enormously. In retrospect we should have introduced versioning of tables much more carefully and with more thought.

Anyway to address it I had to introduce something that I called `TableVersions`. It is a stop gap measure. Its introduction allowed me to caputere the different version of a table (=prerequisite for validation) while still being somewhat backward compatible with existing code. The latter was important as we have a rather organically grown code base that has become very tricky to change without breaking things left and right.

This PR is best reviewed per commit. The commits have rather verbose messages that explain things. It should make reviewing this PR easier. In case you want to check out this branch locally and run it against local schemas and a local meta schema, this is how you do it:

*NOTE*: I've previously changed the version number in the dataset to trigger the validation. The original version numebr dataset matches. The mismatch in table id's in real and actual though.
```console
% schema validate --schema-url=../amsterdam-schema/datasets leidingeninfrastructuur ../amsterdam-schema/schema@v1.1.1/schema.json
Structural validation: success!
Semantic validation:
[ActiveVersionsValidator] Table 'bovengrondsekabels' with version number '1.0.0' does not match with id 'bovengrondseKabels' of the referenced table.

[ActiveVersionsValidator] Table 'ondergrondsekabels' with version number '1.0.0' does not match with id 'ondergrondseKabels' of the referenced table.

[ActiveVersionsValidator] Table 'ondergrondsekabels' with version number '1.2.0' does not match with id 'ondergrondseKabels' of the referenced table.

[ActiveVersionsValidator] Version number '1.2.0' in activeVersions for table 'ondergrondseKabels' does not match with version number '1.0.0' of the referenced table.
```

PS Even though this is a large PR, the actual code for writing the validation is very simple. I really hope we start writing more of them. We definitetly need them. That's for sure.


Yrob:
"However if there are multiple versions dataset["tables"][0] suddenly becomes an entirely different entity, namely a dict that looks like:" Binnen json schema is dit hetzelfde. Volgens json schema moet een reference naar een schema hetzelfde behandeld worden als een inline schema. "When evaluating a $ref, an implementation uses the resolved identifier to retrieve the referenced schema and applies that schema to the instance." Het enige dat anders is is het activeVersions veld dat volgens jsonschema eigenlijk genegeerd zou moeten worden. https://json-schema.org/understanding-json-schema/structuring.html#ref

gkoller:
>  "When evaluating a $ref, an implementation uses the resolved identifier to retrieve the referenced schema and applies that schema to the instance." Het enige dat anders is is het activeVersions veld dat volgens jsonschema eigenlijk genegeerd zou moeten worden. https://json-schema.org/understanding-json-schema/structuring.html#ref

Whoa, that was a major ommission in my understanding of JSON schema. Thank you for clarifying that up. 👍 

But it makes me wonder why we choose to store information, that needs to be retained, in a property that will be lost  once the reference is resolved?

An arguably better design for `dataset["tables"]` would have been an array of arrays (first array the different tables, the second array the table versions). That way we can always keep around all active versions in an Amsterdam Schema compatible way.

src/schematools/types.py
```diff
@@ -37,6 +40,86 @@
 
 ST = TypeVar("ST", bound="SchemaType")
 Json = Union[str, int, float, bool, None, Dict[str, Any], List[Any]]
+Ref = str
+
+
+@total_ordering
+class SemVer:
```
jjmurre: In the amsterdam-schema repo this library is used: https://pypi.org/project/semver/ Could that have been used here as well, or was it not adequate for this use-case?

tests/test_types.py
```diff
@@ -8,17 +8,16 @@
 
 def test_permission_level_ordering() -> None:
     """Test whether enum ordering works based on the int values."""
-    assert sorted(PermissionLevel._member_map_.values()) == [
+    assert sorted(PermissionLevel) == [
```
jjmurre, line 11: 👍🏼 Less clutter, more readable.

src/schematools/db.py
```diff
@@ -49,18 +49,18 @@ def create_meta_table_data(engine, dataset_schema: DatasetSchema):
     session.add(dataset)
 
     for table_data in dataset_schema["tables"]:
-        table_content = {camel_case_to_snake(k): v for k, v in table_data.items() if k != "schema"}
+        table_content = {camel_case_to_snake(k): v for k, v in table_data.default.items() if k != "schema"}
```
jjmurre, line 52: I know this is old code. Maybe we should eat our own dogfood and use our own camel-snake converting functions from `schematools.utils`.

Yrob (commented):
"But I didn't want to contribute
to the mess we are currently in, where the specification does not match
the schema's that in turn do not match the code that operates on them.
Ditching the `activeVersions` information, that we had just regained,
upon serialization to JSON is at least backward compatible with how
DSO-API views things currently." 

I don't understand this with respect to the active versions. How is the implementation inconsistent with the spec? Support for multiple versions is not implemented yet in dso-api, and htere is a note in the spec mentioning this. The activeVersions feature is available in schemas, that does not mean an application using the amsterdam-schema like dso-api has to do anything with it.

src/schematools/types.py
gkoller: Didn't knew that library existed. Could have used that. However it would still involve writing code. As it is, `semver.version.Version` supports additional labels for pre-release and build metadata; we specifically don't. Hence I would need to subclass it to put those restrictions in place.

In addition `SemVer` is-a `str`. Hence I can use it anywhere a `str` is expected without any casts.

So the subclassing + addional casts might amount to the same amount of code (not counting the unit tests). So I'd like to keep it in. But if you are more comfortable with the library I will change things.

jjmurre: After reading the rest of the code, I now see that this `SemVer` implementation is really tailored to our use-case. Maybe we could do it the other way around now, and use our own version in the amsterdam-schema repo :-)

src/schematools/contrib/django/models.py
```diff
@@ -379,6 +385,7 @@ class DatasetTable(models.Model):
 
     dataset = models.ForeignKey(Dataset, on_delete=models.CASCADE, related_name="tables")
     name = models.CharField(max_length=100)
+    version = models.TextField(default=SemVer("1.0.0"))
```
Yrob, line 388: Ik dacht dat in de datasets_dataset en datasets_datasettable tabellen alleen kolommen stonden die dso nodig had. Het zijn toch geen directe representaties van amsterdam-schema objecten?

Yrob (commented):
"This might all change if we ever get a proper Amsterdam Meta Schema
definition for multiple active table versions *inline* with the dataset." 
To json schema there is no distinction between inline and referenced schemas. they represent the same thing. If we want to properly read json schema's we have to properly implement dereffing in schema-tools, not ignore json-schema functionality. Inline tables are unwieldy already.  With multiple versions in the same dataset file it will be worse. This is why table references are best practice in the spec.

gkoller:
> I don't understand this with respect to the active versions. How is the implementation inconsistent with the spec? 

@Yrob I'm sorry for having worded it so negatively, that was inappropriate. I will change it.

Turn out, as you have demonstrated, that the spec was consistent with the meta schema: the `activeVersions` property being discarded is a specified JSON-Schema side-effect of being on the same object as the '$ref`. But what is the value of specifying `activeVersions` and how can we actively mainain and support those versions, if that data is not available after loading the schema definitions? 

> Support for multiple versions is not implemented yet in dso-api, and htere is a note in the spec mentioning this. The activeVersions feature is available in schemas, that does not mean an application using the amsterdam-schema like dso-api has to do anything with it.

What I meant was that given the current Amsterdam Meta Schema, there does not exist a dataset JSON representation that adhers to that meta schema that can capture multiple version of the same table. And as it is, the DSO-API code in `schema-tools` (`schematools.contrib.django`) only accepts a single deserialized JSON document to populate its database tables. It's not that DSO-API does not support multiple tables versions. It's that DSO-API can only consume Amsterdam Schema JSON file and those don't carry informatie about multilpe versions. 

That's what the `TableVersionsEncoder` does; it removes the enriched (not valid Amsterdam Schema) data about multiple table versions so that there is a single version per table

src/schematools/db.py
gkoller: 👍 Will change

gkoller:
> To json schema there is no distinction between inline and referenced schemas. they represent the same thing.

I get that. But due to the fact that `activeVersions` in defined on the same level as the `$ref`, `activeVersions` will be obliterated when the `$ref` is resolved to a single table, removing all knowlegde/information/data about other versions. 

A resolved schema, as the meta schema is currently defined, will only ever have one version per table. There is no way to have multiple versions per tabel inline with the dataset

What we currently have is:

- dataset
   - tables
      - table_1: { table_def }
      - table_2: { table_def }
      - table_3: { table_def }

What I need, but currently is not supported is something akin to;

- dataset
   - tables
      - table_1: [ { table_def_v1 }, { table_def_v2 }, { table_def_v3 }, ... ]
      - table_2: [ { table_def_v1 }, { table_def_v2 }, { table_def_v3 }, ... ]
      - table_3: [ { table_def_v1 }, { table_def_v2 }, { table_def_v3 }, ... ]


> If we want to properly read json schema's we have to properly implement dereffing in schema-tools, not ignore json-schema functionality.

Agreed. But than `activeVersions` should no remain on the same level as the `$ref`.

As I wrote this is as stop gap. Not a definitive solutions. The definitive solution is a rework of tables and version in the Amsterdam Meta Schema. But that would touch even more code. What I was looking for here is a way to retain the information I needed. `TableVersions` looked a good solution. If it worked a next step would be to formalize it is the meta schema, but I needed something working first.

> Inline tables are unwieldy already. With multiple versions in the same dataset file it will be worse. This is why table references are best practice in the spec.

I agree we should support `$ref`s. But after loading the schema, the single JSON document that we end up with should have all information. It currently does not. `activeVersions` is completely gone. I thought that was because we didn't follow the spec. I was wrong. I am sorry about that. But that I know how `$ref` actually work, I think `activeVersions` is modelled incorrectly. It should never have been in a place where it would be obliterated.

src/schematools/contrib/django/models.py
gkoller: It is similar to what Nick did when [he added dataset versioning](https://github.com/Amsterdam/schema-tools/commit/32a47f1fdf50618ea23820afee0f1b88d2f41a66#diff-1b59f231374bf5878b1793280aa14481447515ba99e7f39c5bfeeb6e446898acR199)

jjmurre (approved):
Nice work! Not much to say about the code, looks clean and complete. The underlying problem with the `activeVersions` needs to be adressed. Maybe, now this is all still fresh in you mind, you could write down some thoughts on what would be the best path to fix the `activeVersions` properly.

Yrob:
> 
Ah I misunderstood.
Agreed activeVersions should be moved. We put it where it is so we could implement versioning in amsterdam-schema without breaking dso (after a lot of scope creep on what seemed like a small ticket), and keep everything on the migration path towards supporting multiple versions simultaneously. We now have what we need in amsterdam-schema in terms of content, we just need to make a small change to the structure. But what is a small change in amsterdam-schema, will be a big effort for the rest of the tools.  
tables: [{id:"myTableId", ref:"myTable.json", activeVersions:["1.0.0":"myOldTable","2.0.0":"myTable"]}] should become something along the lines of:  tables:[{id:"myTableId", versions: {"1.0.0": {ref:"myOldTable.json"}, "2.0.0":{ref:"myTable.json"}}, default:"2.0.0" }]

Yrob (approved):

gkoller:
> you could write down some thoughts on what would be the best path to fix the `activeVersions` properly.

Will do.

gkoller:
> Agreed activeVersions should be moved.

👍 

> We put it where it is so we could implement versioning in amsterdam-schema without breaking dso (after a lot of scope creep on what seemed like a small ticket), and keep everything on the migration path towards supporting multiple versions simultaneously. 

That something similar to me introducing `TableVersions` and `TableVersionsEncoder`. It got me what I needed with the least amount of breakage; and still a fair amount of stuff broke.

> We now have what we need in amsterdam-schema in terms of content, we just need to make a small change to the structure. But what is a small change in amsterdam-schema, will be a big effort for the rest of the tools.

Agreed. Potentially a lot will break. I am more than a little hesitant to do that at this stage. But maybe (still very much unsure) addressing it now is the better path forward. We need to discuss that next week.

src/schematools/types.py
gkoller: I might extract it into `more-ds`, but will do so in a separate PR.

gkoller:
I have changed some of the commit messages and comments in the code to not be so negative. As said before that was inappropriate.

PR #222, opened 2021-11-25
Reformat docs/import_event.py

lbam:
This pops up in `git grep` sometimes, making its output hard to scroll through.


jjmurre (approved):
LGTM, easier on the cli.

PR #224, opened 2021-11-29
Guido/fix from dict

gkoller:
A deepcopy is required if we use a `dict` as the basis of our datastructures. Afterall we don't want our internal representation to be altered because the outside world still had a reference to the `dict` that was originally passed in to `from_dict`

Extracted `SemVer` into `more_ds`

### DIFF
--- a/src/schematools/cli.py
+++ b/src/schematools/cli.py
@@ -10,6 +10,7 @@ import requests
 import sqlalchemy
 from deepdiff import DeepDiff
 from json_encoder import json
+from jsonschema import draft7_format_checker
 from sqlalchemy import create_engine
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.schema import CreateTable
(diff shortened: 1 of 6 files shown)


src/schematools/cli.py
```diff
@@ -363,7 +364,9 @@ def validate(
     structural_errors = False
     try:
         click.echo("Structural validation: ", nl=False)
-        jsonschema.validate(instance=dataset.json_data(), schema=meta_schema)
+        jsonschema.validate(
```
jjmurre, line 367: Nice addition!

jjmurre (approved):
LGTM. Nice that SemVer is now widely available.

lbam:
I'd like to see a test for the deepcopy.

gkoller:
> I'd like to see a test for the deepcopy.

Will add test.

gkoller:
@lbam The test proofed valuable, as it turns out de deepcopy doesn't do anything valuable here. Reading up on `UserDict` (a base class of `DatasetSchema`)  I see that:

>  The instance’s contents are kept in a regular dictionary, which is accessible via the data attribute of UserDict instances. If initialdata is provided, data is initialized with its contents; note that a reference to initialdata will not be kept, allowing it to be used for other purposes.

So a deepcopy is unnecessary. I gues I have to look at another fix for the failing tests in `dso-api`.

lbam:
Maybe push the test without the changes?

PR #228, opened 2021-12-01
Avoid adding compound keys to through table when relation is loose

DonQueso89:
Since we are only pointing to the primary identifier of the target relation, we only need that FK in the through table.

### DIFF
--- a/tests/django/test_models.py
+++ b/tests/django/test_models.py
@@ -240,6 +240,11 @@ def test_model_factory_loose_relations(meldingen_dataset, gebieden_dataset):
 def test_model_factory_loose_relations_n_m_temporeel(woningbouwplannen_dataset, gebieden_dataset):
     """Prove that a loose relation is created when column
     is part of relation definition (<dataset>:<table>:column)
+    and that the intermediate model contains the correct references to both
+    associated tables.
+
+    Loose m2m relations defined with an array of scalars or an array of
+    single-property-objects generate the same output.
     """
     model_dict = {
         cls._meta.model_name: cls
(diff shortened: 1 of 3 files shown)


tests/django/test_models.py
```diff
@@ -250,12 +255,27 @@ def test_model_factory_loose_relations_n_m_temporeel(woningbouwplannen_dataset,
...
+    intermediate_table = buurten_field.remote_field.through
     assert isinstance(buurten_field, LooseRelationManyToManyField)
-    assert isinstance(buurten_field.remote_field.through, ModelBase)
+    assert isinstance(intermediate_table, ModelBase)
+
+    assert sorted([x.name for x in intermediate_table._meta.get_fields()]) == [
+        "buurten",
+        "id",
+        "woningbouwplan",
+    ]
```
gkoller, line 266: No need for a generator to create a list and sort it afterwards, use sets instead
```python
assert {x.name for x in intermediate_table._meta.get_fields() } == { "buurten", "id", "woningbouwplan"}
```

tests/django/test_models.py
```diff
@@ -250,12 +255,27 @@ def test_model_factory_loose_relations_n_m_temporeel(woningbouwplannen_dataset,
...
+
     buurten_as_scalar_field = meta.get_field("buurten_as_scalar")
+    intermediate_table = buurten_as_scalar_field.remote_field.through
     assert isinstance(buurten_as_scalar_field, LooseRelationManyToManyField)
     assert isinstance(buurten_as_scalar_field.remote_field.through, ModelBase)
 
+    assert sorted([x.name for x in intermediate_table._meta.get_fields()]) == [
```
gkoller, line 273: See https://github.com/Amsterdam/schema-tools/pull/228/files#r760513815

tests/django/test_models.py
```diff
@@ -250,12 +255,27 @@ def test_model_factory_loose_relations_n_m_temporeel(woningbouwplannen_dataset,
...
+        "buurten_as_scalar",
+        "id",
+        "woningbouwplan",
+    ]
```
lbam, line 277:
```suggestion
    assert {x.name for x in intermediate_table._meta.get_fields()]} == {"buurten_as_scalar", "id", "woningbouwplan"}
```

jjmurre (approved):
LGTM, nice catch!

PR #230, opened 2021-12-13
Relax flake8 checks in pre-commit hook

lbam:
We use print statements in too many places to fix. We also use the names of builtins for fields and parameter names. Changing what would mean changing our public API.


gkoller (approved):
I'm sad to see them go, but I do understand it is better that way.

PR #236, opened 2022-01-13
Clarify id vs. name in DatasetFieldSchema docstrings

lbam:
Also note that `has_shortname` shouldn't really be used. Applications should just get the name.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -1036,22 +1036,7 @@ class DatasetTableSchema(SchemaType):
 
 
 class DatasetFieldSchema(DatasetType):
-    """A single field (column) in a table
-
-    This class has an `id` property to uniquely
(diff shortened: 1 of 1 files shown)


src/schematools/types.py
```diff
@@ -1092,11 +1078,22 @@ def is_autoincrement(self) -> bool:
...
     @property
     def name(self) -> str:
-        """Table name, for display purposes only."""
+        """
+        The name of a field is used to derive its column name in SQL.
+
+        The name is equal to the id, except when it is overridden by the presence
+        of a shortname.
+
+        The actual column name in SQL is the snake-casing of the name.
```
jjmurre, line 1087: Good to add this, to avoid confusion!

jjmurre (approved):
LGTM

PR #238, opened 2022-01-17
Add docstring for DatasetFieldSchema.type

lbam:


src/schematools/types.py
```diff
@@ -1093,6 +1093,12 @@ def required(self) -> bool:
 
     @property
     def type(self) -> str:
+        """Returns the type of this field.
```
jjmurre, line 1096: Maybe you can also point to the documentation at: https://schemas.data.amsterdam.nl/docs/ams-schema-spec.html#data-types

jjmurre (approved):
LGTM

src/schematools/types.py
lbam: Done, please check again.

jjmurre (approved):
LGTM

PR #239, opened 2022-01-17
Improvements for M2M relationships, fix joining loose relations

vdboor:
* Improved tracking reverse relations (by changing the reverse names).
* Improved __repr__ for clarity.
* Fixed the `to_field` for the right-side of loose m2m relations

### DIFF
--- a/src/schematools/contrib/django/factories.py
+++ b/src/schematools/contrib/django/factories.py
@@ -180,6 +180,23 @@ class FKRelationMaker(RelationMaker):
 
         if self.field.is_composite_key:
             kwargs["to_fields"] = [to_snake_case(field.id) for field in self.field.subfields]
+        elif (
+            # HACK: This complicated logic is needed because self.field.is_loose_relation
+            # has very mixed-up logic that handles things which the callers should have handled.
+            # This makes it impossible to determine whether a through-table has loose relations.
+            # Solving that turns out to be really complex and bring up more issues. However,
+            # reading the NM-field does work, so at least one side of the relation can be fixed.
+            (nm_field := self.field.table.parent_table_field) is not None
+            and nm_field.is_loose_relation
+            and self.field.id == self.field.table["throughFields"][1]
+        ):
+            # A FK in the through table might actually be a partial composite key.
+            # When this is the case, make sure the to_field points to the right field.
+            target_field_ids = nm_field.related_field_ids
+            target_field = self.field.related_table.get_field_by_id(target_field_ids[0])
+
+            if target_field_ids[0] != "id" and not target_field.is_primary:
+                kwargs["to_field"] = to_snake_case(target_field_ids[0])
 
         return kwargs
 
--- a/src/schematools/contrib/django/factories.py
+++ b/src/schematools/contrib/django/factories.py
@@ -131,17 +131,40 @@ class FKRelationMaker(RelationMaker):
         If the linked table describes the other end of the relationship,
         this field will also be included in the model.
         """
-        if self.field._parent_table.has_parent_table:
+        parent_table = self.field.table
+        if parent_table.is_nested_table:
             # Won't ever show related name for internal tables
-            return to_snake_case(self.field._parent_table["originalID"])
-        elif self.field._parent_table.is_through_table:
-            # Need this for walking over the through table for the "_links" section.
-            return to_snake_case(f"{self.field._parent_table.id}_through_{self.field.id}")
+            return to_snake_case(parent_table["originalID"])
+        elif parent_table.is_through_table:
+            # This provides a reverse-link from each FK in the M2M table to the linked tables,
+            # allowing to walk *inside* the M2M table instead of over it.
+            # For debugging purposes these names are clarified depending on what direction
+            # the key takes.
+            m2m_field = parent_table.parent_table_field
+            through_fields = parent_table["throughFields"]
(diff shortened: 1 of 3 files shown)


src/schematools/types.py
```diff
@@ -714,6 +714,12 @@ def __init__(
 
     def __repr__(self):
         return f"<{self.__class__.__name__}: {self['id']}>"
+        prefix = ""
+        if self._parent_schema is not None:
+            prefix += f"{self._parent_schema.id}."
+        if self._parent_table is not None:
+            prefix += f"{self._parent_table.id}."
+        return f"<{self.__class__.__name__}: {prefix}{self['id']}>"
```
lbam, line 722: You forgot to remove the original implementation, above.

src/schematools/types.py
```diff
@@ -1139,6 +1159,21 @@ def related_table(self) -> Optional[DatasetTableSchema]:
             related_table_id, include_nested=False, include_through=False
         )
 
+    @property
+    def related_field_ids(self) -> Optional[list[str]]:
+        """IF this field is a relation, return which fields this relation references.
```
lbam, line 1164:
```suggestion
        """If this field is a relation, return which fields this relation references.
```

src/schematools/contrib/django/factories.py
```diff
@@ -131,17 +131,40 @@ def _get_related_name(self):
...
+                if m2m_reverse_name is not None:
+                    # As the field exists on the second model,
+                    # let the reverse relation also reflect that.
+                    return f"rev_m2m_{to_snake_case(m2m_reverse_name.id)}"
+
+            # By default, create something unique and recognizable.
+            # The "m2m" would be snake_cased as m_2_m, so it's added afterwards.
+            # The parent table ID already has the main field name included, but for tables
+            # with a self-reference, the field is still added to guarantee uniqueness.
+            return "rev_m2m_" + to_snake_case(f"{parent_table.id}_via_{self.field.id}")
```
lbam, line 160: Shouldn't this use `self.field.name` instead of the `id`?

src/schematools/types.py
vdboor: yikes. debugging mistakes..

src/schematools/types.py
vdboor: tx.

src/schematools/contrib/django/factories.py
```diff
@@ -131,17 +131,40 @@ def _get_related_name(self):
...
+            if self.field.name == through_fields[0]:
+                # First model, can resemble the original field name,
+                return f"rev_m2m_{to_snake_case(m2m_field.name)}"
+            elif self.field.name == through_fields[1]:
+                # Second model, can resemble the reverse name if it exists.
+                m2m_reverse_name = m2m_field.reverse_relation
+                if m2m_reverse_name is not None:
+                    # As the field exists on the second model,
+                    # let the reverse relation also reflect that.
+                    return f"rev_m2m_{to_snake_case(m2m_reverse_name.id)}"
```
lbam, line 154: Here too: `.id` or `.name`?

src/schematools/contrib/django/factories.py
vdboor: nope :-) The `.name` is used for the `shortname` in SQL. All Django model field names use the long identifiers, so this reverse name is kept in line with that to clearly see which field it's opposite end is.

src/schematools/contrib/django/factories.py
vdboor: same as above!

src/schematools/types.py
```diff
@@ -748,6 +748,14 @@ def dataset(self) -> DatasetSchema:
         """The dataset that this table is part of."""
         return self.parent_schema
 
+    @property
+    def parent_table_field(self) -> Optional[DatasetFieldSchema]:
+        """Provide the NM-relation that generated this through table."""
+        if self.through_table or self.nested_table:
+            return self._parent_table.get_field_by_id(self["originalID"])
```
jjmurre, line 760: Be carefull to check on this. @gkoller changed `_parent_table` to `parent_table`.

src/schematools/contrib/django/factories.py
```diff
@@ -131,17 +131,40 @@ def _get_related_name(self):
...
-        elif self.field._parent_table.is_through_table:
-            # Need this for walking over the through table for the "_links" section.
-            return to_snake_case(f"{self.field._parent_table.id}_through_{self.field.id}")
+            return to_snake_case(parent_table["originalID"])
+        elif parent_table.is_through_table:
+            # This provides a reverse-link from each FK in the M2M table to the linked tables,
+            # allowing to walk *inside* the M2M table instead of over it.
+            # For debugging purposes these names are clarified depending on what direction
+            # the key takes.
+            m2m_field = parent_table.parent_table_field
```
jjmurre, line 143: Ouch, this took some serious Django relations code-reading.

src/schematools/types.py
vdboor: Yes, this checks the internal field, which is fine. But the outside world should use the public property off course.

src/schematools/contrib/django/factories.py
vdboor: indeed.. ☺️

jjmurre (approved):
LGTM

PR #242, opened 2022-01-19
Revert changes to fix broken _parent_schema field.

vdboor:
Does what @gkoller intended, without breaking the other changes from #239. Can check later whether the attribute naming should be improved.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -517,7 +517,7 @@ class DatasetSchema(SchemaType):
                 table.name, snakecased_fieldname
             )
         return DatasetTableSchema(
-            sub_table_schema, parent_schema=self, parent_table=table, nested_table=True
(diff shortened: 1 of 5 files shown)


src/schematools/types.py
```diff
@@ -715,14 +715,14 @@ def __init__(
         self,
         *args: Any,
         parent_schema: DatasetSchema,
-        parent_table: Optional[DatasetTableSchema] = None,
+        _parent_table: Optional[DatasetTableSchema] = None,
```
lbam, line 718: An argument with a leading `_` looks really strange to me.

jjmurre (approved):
LGTM

src/schematools/types.py
vdboor: yes this was designed as an internal data-passing. It might need changes, but first I'd like to revert things that broke stuff.

lbam: Ok.

PR #245, opened 2022-01-20
Downgrade jsonschema

jjmurre:
See commit message.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = amsterdam-schema-tools
-version = 3.3.2
+version = 3.3.3
 url = https://github.com/amsterdam/schema-tools
 license = Mozilla Public 2.0
 author = Amsterdam Data en Informatie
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -32,7 +32,7 @@ install_requires =
     pg-grant
     click
     deepdiff
-    jsonschema[format] >= 4.2.1
+    jsonschema[format] ~= 3.2.0
```
lbam, line 35: I would have expected `>=`. What does `~=` do?

jjmurre: Using >= 3.2.0 would select version 4.2.1. This compatible release spec (https://www.python.org/dev/peps/pep-0440/#compatible-release) keeps the major version the same, so, no major version = 4 would be installed.

I could also be `~= 3.0.0`, because `~= 3.2.0` will also preserve the minor version. In this case this does not matter, because 3.2.0 is the latest version in the 3 range.  However, maybe a bugfix release in the 3-range could be release, so thinking of it `~= 3.0.0` would be a better choice.

lbam (approved):

PR #247, opened 2022-01-24
Add CompositeForeignKeySubField that handles relations to temporal field keys.

vdboor:
This allows the `to_field` to point to non-unique fields.
Additional temporal query filters ensure the result is still unique.


src/schematools/contrib/django/factories.py
```diff
@@ -167,20 +170,13 @@ def _get_related_name(self):
...
-        # In schema foreign keys should be specified without _id,
-        # but the db_column should be with _id
-        kwargs = {
-            **super().field_kwargs,
-            "db_column": f"{to_snake_case(self.field.name)}_id",
-            "db_constraint": False,
-            "related_name": self._get_related_name(),
-        }
+    def _get_to_field(self) -> Optional[str]:
+        """Tell whether the foreign key doesn't point to the targets's primary key.
```
lbam, line 174: This suggests that this a predicate, but the return value is a string.

src/schematools/contrib/django/factories.py
```diff
@@ -196,7 +192,26 @@ def field_kwargs(self):
...
+
+        return None
+
+    @property
+    def field_kwargs(self):
+        # In schema foreign keys should be specified without _id,
+        # but the db_column should be with _id
+        kwargs = {
+            **super().field_kwargs,
+            "db_column": f"{to_snake_case(self.field.name)}_id",
```
lbam, line 205:
```suggestion
            "db_column": to_snake_case(self.field.name) + "_id",
```

Easier to read, IMHO.

src/schematools/contrib/django/factories.py
vdboor: ok, updated!

src/schematools/contrib/django/factories.py
vdboor: good point. This code wasn't changed, will check later (more PRs coming)

lbam (approved):

PR #248, opened 2022-01-25
AB#27817 Consider shortnames when applying table perms

DonQueso89:
While implementing this, I noticed that the exception handler inside the SQL statement was silently swallowing errors whenever we try to run a GRANT on a non-existing field or table. Removing this revealed two other bugs:

1) applying permissions to FK fields doesnt work because the name used to apply the perm does not include the `_id` suffix
2) snakecasing dataset ids breaks the command because dataset ids are imported with their raw id (i.e. unsnakecased)

The first problem is now explicitly omitted with a FIXME and ive added a [ticket](https://dev.azure.com/CloudCompetenceCenter/Data%20Diensten/_backlogs/backlog/Data%20Diensten%20Team/Stories/?workitem=28436)
The second problem has been solved by #249

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,27 @@
+# This compose file is supplied for local dev
+# Since there is no running process, it just starts a bash
+# session in order to run tests etc.
+version: "3.0"
+services:
+  database:
+    image: amsterdam/postgres11
+    environment:
+      POSTGRES_DB: dataservices
+      POSTGRES_USER: dataservices
+      POSTGRES_PASSWORD: insecure
+
+  app:
+    build: .
+    links:
+      - database
+    environment:
+      SECRET_KEY: insecure
+      SCHEMA_URL: https://schemas.data.amsterdam.nl/datasets/
+      DATABASE_URL: "postgres://dataservices:insecure@database/dataservices"
+    volumes:
+      - ".:/app"
+      # Uncomment (and adjust the location of) this mount to experiment with
+      # edited datasets from your local amsterdam-schema clone
+      # - "../amsterdam-schema/datasets:/tmp/datasets"
+    command: >
+      bash
--- a/src/schematools/permissions/db.py
+++ b/src/schematools/permissions/db.py
@@ -12,6 +12,17 @@ PUBLIC_SCOPE = "OPENBAAR"
 existing_roles = set()
 
 
+def is_remote(table_name):
+    """WARNING: UGLY HACK until 265311 is resolved and we
+    can interrogate schematools to find out whether a table
+    is remote."""
+    return (
+        table_name.startswith("haalcentraalbrk")
+        or table_name.startswith("haalcentraalbag")
+        or table_name.startswith("brp")
+    )
+
+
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    links:
+      - database
+    environment:
+      SECRET_KEY: insecure
+      SCHEMA_URL: https://schemas.data.amsterdam.nl/datasets/
+      DATABASE_URL: "postgres://dataservices:insecure@database/dataservices"
+    volumes:
+      - ".:/app"
+    command: >
+      bash
```
DonQueso89, line 21: added this for local development

src/schematools/permissions/db.py
```diff
@@ -393,16 +416,12 @@ def _execute_grant(session, grant_statement, echo=True, dry_run=False):
     We don't want to break out of the session just because a table, column, or user doesn't
```
jjmurre, line 416: I guess this comment is not adequate anymore, right?

src/schematools/permissions/db.py
```diff
@@ -393,16 +416,12 @@ def _execute_grant(session, grant_statement, echo=True, dry_run=False):
...
     exist.
     """
-
     status_msg = "Skipped" if dry_run else "Executed"
     sql_statement = f"""
```
jjmurre, line 420: The airflow grants operator is calling in to schematools from here: https://github.com/Amsterdam/dataservices-airflow/blob/7d9db2e427e79c622a3bd5479f415bf2dd69e1d5/src/plugins/postgres_permissions_operator.py#L163

Could this break, after removing the postgres exception handler?

jjmurre: If needed, I can do some testing for you, I have a working development Airflow.

docker-compose.yml
lbam: Maybe put that in a comment at the top?

src/schematools/permissions/db.py
```diff
@@ -12,6 +12,19 @@
...
+    """WARNING: UGLY HACK until 265311 is resolved and we
+    can interrogate schematools to find out whether a table
+    is remote."""
+    if (
+        table_name.startswith("haalcentraalbrk")
+        or table_name.startswith("haalcentraalbag")
+        or table_name.startswith("brp")
+    ):
+        return True
+    return False
```
lbam, line 25:
```suggestion
    return (
        table_name.startswith("haalcentraalbrk")
        or table_name.startswith("haalcentraalbag")
        or table_name.startswith("brp")
    )
```

jjmurre (dismissed):
LGTM. Can you make a ticket in ADO for Iteration 5 to check for the removal of the ugly hack in the `is_remote` function.

src/schematools/permissions/db.py
DonQueso89: that depends on the versions of the schemas that it is called with. Ive run it against the latest versions without error

jjmurre (approved):
LGTM

PR #251, opened 2022-01-26
Drop tbump

lbam:
tbump wants to push to master on GH, which we don't allow anymore.
Replace it by instructions for doing a release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,14 +97,20 @@ However not all `json` files are Amsterdam schema files.
 To exclude files or directories use `exclude` with pattern.
 
 `pre-commit` depends on properly tagged revisions of its hooks.
-Hence we should take care to, not only bump version numbers
-on updates to this package,
-but also commit a tag with the version number.
-This is automated by means of the `tbump` tool.
-Bumping a version from 0.18.1 to 0.18.2
-and generating the appropriate git commits/tags
-is as easy as running:
+Hence, we should not only bump version numbers on updates to this package,
+but also commit a tag with the version number; see below.
 
-```console
-$ tbump 0.18.2
-```
+## Doing a release
+
+(This is for schema-tools developers.)
+
+We use GitHub pull requests. If your PR should produce a new release of
+schema-tools, make sure one of the commit increments the version number in
+``setup.cfg`` appropriately. Then,
+
+* merge the commit in GitHub, after review;
+* pull the code from GitHub and merge it into the master branch,
+  ``git checkout master && git fetch origin && git merge --ff-only origin/master``;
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -97,14 +97,20 @@ However not all `json` files are Amsterdam schema files.
...
+
+(This is for schema-tools developers.)
+
+We use GitHub pull requests. If your PR should produce a new release of
+schema-tools, make sure one of the commit increments the version number in
+``setup.cfg`` appropriately. Then,
+
+* merge the commit in GitHub, after review;
+* pull the code from GitHub and merge it into the master branch,
+  ``git checkout master && git fetch origin && git merge origin/master``;
```
vdboor, line 113:
```suggestion
  ``git checkout master && git pull --rebase``;
```

or: `git merge origin/master --no-ff`

README.md
```diff
@@ -97,14 +97,20 @@ However not all `json` files are Amsterdam schema files.
...
+* tag the release X.Y.Z with ``git tag vX.Y.Z``;
```
vdboor, line 114:
```suggestion
* tag the release X.Y.Z with ``git tag vX.Y.Z -m "message"``;
```

(or `git tag -a`). tbump generated annotated commits which includes dates+author names.
Without the annotation the tag is only a pointer like a branch name is.

README.md
```diff
@@ -97,14 +97,20 @@ However not all `json` files are Amsterdam schema files.
...
+* push the tag to GitHub with ``git push origin --tags``;
+* release to PyPI (requires the PyPI secret).
```
vdboor, line 116: explain how to release. e.g. `make build upload`

README.md
lbam: `pull --rebase` is dangerous since it may put the tag on a local commit.

`--no-ff` is worse because always creates a merge commit and the tag will end up on that commit. Did you mean `--ff`?

README.md
lbam: Oh joy, the multiple types of Git tags... does it matter in this case? We just want a specific commit to be picked up by PyPI, right? (I'm sure I did this wrong with the last release.)

README.md
lbam: Done. `upload` requires `build`, so `make upload` should be enough,

README.md
lbam: Ok, fixed now. Looks like I only screwed up one tag.

README.md
lbam: @vdboor Made it `--ff-only`.

jjmurre (approved):
LGTM

PR #252, opened 2022-01-27
Rewrite schema validators as decorated functions

lbam:
All decorators were stateless classes with a single validate method that
were attached to their base class. This made writing new validators a
pain because of all the boilerplate and the flake8 requirement of having
a docstring both on the class and on the validate method.

The new code uses ordinary standalone functions with a decorator to
register them as validators. They're now also all marked as private with
the familiar underscore, and they have a human-readable name that is
independent of their identifier.

### DIFF
--- a/src/schematools/validation.py
+++ b/src/schematools/validation.py
@@ -16,37 +16,15 @@ Structural validation cannot verify that the properties specified in that array
 actually do exist in the schema.
 Semantic validation can and should.
 
-Specific semantic validations are created by subclassing :class:`Validator`
-and overriding the :meth:`Validator.validate` method.
-Simply by means of subclassing,
-the new validator is automatically registered.
-Registered validator classes will all be run
-when :meth:`Validator.run_all()` is invoked.
-
-.. note::
-
-   For the registration to work
-   all :class:`Validator` subclasses need to be parsed by the Python interpreter.
-   This can be achieved by importing the module they reside in.
-
-Example:
-    The following will run all registered validators
-    on a dataset ``dataset``.
-    Any validation errors are printed to ``stdout``::
-
-        dataset = _get_dataset_schema(meta_schema_url, schema_location)
-        validator = Validator(dataset=dataset)
-        for error in validator.run_all():
-            print(error)
-
+Semantic validations are created decorating a function with `_validator`.
 """
 from __future__ import annotations
 
 import operator
 import re
 from dataclasses import dataclass
-from functools import partial
-from typing import Callable, ClassVar, Iterator, List, Optional, Set, Type, cast, final
+from functools import partial, wraps
+from typing import Callable, Iterator, List, Optional, Set, cast
 
 from schematools import MAX_TABLE_NAME_LENGTH
 from schematools.types import DatasetSchema, SemVer, TableVersions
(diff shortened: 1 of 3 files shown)


jjmurre:
I must say, I like this simplification!

src/schematools/validation.py
```diff
@@ -79,192 +57,166 @@ def __init__(self, message: str) -> None:  # noqa: D107
...
+            yield ValidationError(validator_name=name, message=msg)
 
-    def __init__(self, dataset: DatasetSchema) -> None:
-        """Initialize the validator with a dataset.
 
-        Args:
-            dataset: The dataset to run the validations on.
-        """
-        self.dataset = dataset
+def _validator(name: str) -> Callable:
```
jjmurre, line 75: What's the rationale behind the starting `_`?

lbam: Privacy. I'm not sure if this is considered a public module at all, but there are currently no validators outside this file. We can always remove the underscore if we want to support validators from user code (but why would we?).

jjmurre: Yeah, makes sense.

jjmurre (approved):
LGTM, less boilerplate.

src/schematools/validation.py
vdboor, line 75: perhaps call it `_register_validator()`? That's a bit more descriptive of what it does

src/schematools/validation.py
```diff
@@ -79,192 +57,166 @@ def __init__(self, message: str) -> None:  # noqa: D107
...
-                # I think this is a bug is schema-tools, but for now I'll cover this case
-                # explicitly.
-                remove_id_suffix = cast(Callable[[str], str], partial(re.sub, r"(.+)Id", r"\1"))
-                derived_fields = tuple(
-                    map(
-                        lambda f: DerivedField(original=remove_id_suffix(f), derived=f),
-                        missing_fields,
-                    )
+    @dataclass
+    class DerivedField:
```
vdboor, line 140: Use `NamedTuple` instead?

lbam: This is from Guido's code.

tests/test_validation.py
```diff
@@ -10,63 +10,65 @@
...
-    ActiveVersionsValidator,
-    IdentPropRefsValidator,
-    PsqlIdentifierLengthValidator,
-    ValidationError,
-    _validate_camelcase,
+    _active_versions,
+    _camelcase_ident,
+    _identifier_properties,
+    _postgres_identifier_length,
+    run_all,
```
vdboor, line 17: perhaps `from schematools import validation` so it's more obvious that when calling those private functions it's ovious they don't sit in this file.

src/schematools/validation.py
lbam: Done.

vdboor:
I really like this simplification too. You can choose to prefix all functions with `_validate_...()` but I'm not sure that helps.

tests/test_validation.py
lbam: I have an even better idea. I changed the tests to use `run_all` (renamed `run`; maybe just call it `validate`?) instead of the specific private functions. That should give better test coverage.

lbam:
I thought of prefixing everything, but the decorator should make it clear that the functions validate. The name of the module is a hint too ;)

jjmurre (approved):
LGTM

PR #253, opened 2022-01-31
[FIX] camelCase in dataset id

cvriel:
The `dataset id` is used for defining the database schema where the tables of that dataset will reaside. In some cases the dataset id is camelCased. Leading to also the prefix of tables - that contain the name of its schema - to have camelCased prefixes. Which leads to wrong table names and issues for the datateams. To cope with this, the dataset id is snaked cased.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = amsterdam-schema-tools
-version = 3.3.5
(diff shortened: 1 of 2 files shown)


jjmurre (approved):
LGTM. Nice catch!

jjmurre (approved):
LGTM

PR #256, opened 2022-02-02
[AB#29149] Identifier fields should not have separate "auth" scopes

lbam:


jjmurre (approved):
LGTM, met de nieuwe validator aanpak!

PR #264, opened 2022-02-21
[AB#24838] Add get_field_by_id to DatasetFieldSchema

lbam:
The name of this method matches that of the one on DatasetTableSchema, rather than those of the other DatasetFieldSchema methods (`get_subfields` etc.). That's on purpose: the filter validation in DSO-API assumes it can use the same method on both types to walk a schema and resolve a compound field name.


jjmurre (approved):
LGTM

PR #268, opened 2022-03-23
Disallow relations pointing to authorized fields

lbam:
We have no rule defining how these are supposed to work. The patch disallows them outright.


tests/test_validation.py
```diff
@@ -107,3 +107,10 @@ def test_main_geometry(here: Path) -> None:
     assert error.message == (
         "mainGeometry = 'merkOmschrijving' is not a geometry field, type = 'string'"
     )
+
+
+def test_rel_auth(here: Path) -> None:
+    dataset = dataset_schema_from_path(here / "files" / "rel_auth.json")
```
jjmurre, line 113: This does not test all the scenarios for `auth`. 
Maybe it is easier to read in the json for the dataset into a dict and tweak it on the fly and then use `DatasetSchema.from_dict`, instead of writing out all datasets in json format.

tests/files/rel_auth.json
```diff
@@ -0,0 +1,59 @@
...
+  "id": "relauth",
+  "title": "Relations with auth",
+  "status": "beschikbaar",
+  "version": "1.2.3",
+  "tables": [
+    {
+      "id": "base",
+      "type": "table",
+      "version": "1.2.3",
+      "auth": ["SCOPE/HAMMERTIME"],
```
jjmurre, line 12: I like the names :-)

lbam: Finally, someone notices. Amsterdam/dso-api#417

tests/test_validation.py
lbam: Done.

jjmurre (approved):
LGTM.

PR #270, opened 2022-03-29
Remove PermissionCollection and custom control flow

lbam:
We don't need to build up a list of permissions in the functions that use `PermissionCollection`: we only ever use the highest permission we find, so track that. Also, the abort_on_highest decorator is "DRY" but much more complicated than just checking if we've found `PermissionLevel.highest` at the start of the loop.


vdboor:
nice!

vdboor (approved):

PR #275, opened 2022-04-05
Fix permissions.db

lbam:
This too was broken by the `scopes=[]` -> `scopes={"OPENBAAR"}` change.


jjmurre (approved):
LGTM. Nice, I was making exactly the same changes :-) As said, we have to look deeper into dependencies between scopes.

PR #282, opened 2022-05-04
Take PUBLIC_SCOPE into account during validation

jjmurre:
Recently, a public scope "OPENBAAR" has been added to all datasets.
The validator that validates the scopes, should take this PUBLIC_SCOPE
into account.

### DIFF
--- a/src/schematools/permissions/__init__.py
+++ b/src/schematools/permissions/__init__.py
@@ -1,3 +1,4 @@
+from .auth import PUBLIC_SCOPE  # noqa: F401, D401
 from .auth import Permission, UserScopes
 
 __all__ = (
--- a/src/schematools/permissions/db.py
+++ b/src/schematools/permissions/db.py
@@ -9,14 +9,13 @@ from sqlalchemy import text
 from sqlalchemy.engine import Engine
 from sqlalchemy.orm import Session, sessionmaker
 
+from schematools.permissions import PUBLIC_SCOPE
 from schematools.types import DatasetSchema
 from schematools.utils import to_snake_case
 
 # Create a module-level logger, so calling code can
 # configure the logger, if needed.
 logger = logging.getLogger(__name__)
-
-PUBLIC_SCOPE = "OPENBAAR"
 
 existing_roles = set()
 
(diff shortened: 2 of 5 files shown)


src/schematools/permissions/db.py
```diff
@@ -9,14 +9,15 @@
...
+from schematools.permissions import PUBLIC_SCOPE
 from schematools.types import DatasetSchema
 from schematools.utils import to_snake_case
 
 # Create a module-level logger, so calling code can
 # configure the logger, if needed.
 logger = logging.getLogger(__name__)
 
-PUBLIC_SCOPE = "OPENBAAR"
+# PUBLIC_SCOPE = "OPENBAAR"
```
lbam, line 20:
```suggestion
```

jjmurre: Zag-em ook net :-)

src/schematools/permissions/db.py
```diff
@@ -9,15 +9,14 @@
 from sqlalchemy.engine import Engine
 from sqlalchemy.orm import Session, sessionmaker
 
+from schematools.permissions import PUBLIC_SCOPE
```
lbam, line 12: Doesn't this lead to an import cycle? It did the last time I tried removing the constant from this file.

lbam (approved):

lbam:
LGTM modulo the potential import cycle. If you can run this locally, feel free to merge.

src/schematools/permissions/__init__.py
```diff
@@ -1,3 +1,5 @@
+"""Base package for permissions."""
```
lbam, line 1: Side remark, but this doc isn't very informative. If this is to satisfy a checker, maybe we should get set it to a laxer mode.

jjmurre: Agree.

src/schematools/permissions/db.py
jjmurre: When running the tests, this module is touched, but I don't get an import cycle. Is there a way to trigger it?

src/schematools/permissions/__init__.py
jjmurre: I laxed this flake8 check in setup.cfg.

src/schematools/permissions/db.py
lbam: Should be ok then.

lbam (approved):

PR #288, opened 2022-05-30
Add bandit security linter to pre-commit hooks

lbam:
For [AB#46681](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/46681). Also fixed some errors that it found and added a "nosec" annotation to the final one: a jinja2.Environment was created with HTML/XML autoescaping off. We're not generating XML here, though.


jjmurre (approved):
LGTM

PR #289, opened 2022-06-02
Add DatasetTableSchema.get_additional_relation_by_id()

vdboor:
This is needed for the filter logic to support reverse relations

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -836,6 +836,17 @@ class DatasetTableSchema(SchemaType):
 
         raise SchemaObjectNotFound(f"Field '{field_id}' does not exist in table '{self.id}'.")
 
+    @lru_cache()  # type: ignore[misc]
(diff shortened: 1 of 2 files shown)


src/schematools/types.py
```diff
@@ -836,6 +836,17 @@ def get_field_by_id(self, field_id: str) -> DatasetFieldSchema:
 
         raise SchemaObjectNotFound(f"Field '{field_id}' does not exist in table '{self.id}'.")
 
+    @lru_cache()  # type: ignore[misc]
+    def get_additional_relation_by_id(self, relation_id: str) -> AdditionalRelationSchema:
+        """Get a fields based on the ids of the field."""
```
lbam, line 841: Docstring doesn't match the function.

vdboor: tx!

jjmurre (approved):
LGTM

PR #307, opened 2022-07-27
Remove deprecated and unused code

lbam:
None of this was used in DSO-API or dataservices-airflow. `get_fk_fields` was used internally, so it's now a private method. It also returns `[DatasetFieldSchema]` instead of `[str]` to be more generally useful and to allow reuse of the `db_name` method (-1 `to_snake_case`!). `schema_fetch_url_file` was moved to its only spot of use and made private, with the previous deprecation method as a comment.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = amsterdam-schema-tools
-version = 3.6.11
+version = 4.0.0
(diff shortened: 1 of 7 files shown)


vdboor (approved):

vdboor:
mooie opschoning!

PR #308, opened 2022-07-28
AB#50108: Convert LooseRelationField into an actual relation

vdboor:
This changes the LooseRelationField into an actual foreign key, so applications can perform joins on the relation. This change is backwards incompatible, and DSO-API will have an corresponding PR to be compatible with this change.

This only affects the fields that are *not* part of a composite foreign key. Those remain simple charfields.

### DIFF
--- a/src/schematools/contrib/django/models.py
+++ b/src/schematools/contrib/django/models.py
@@ -549,26 +549,28 @@ class Profile(models.Model):
         return self.name
 
 
-class LooseRelationField(models.CharField):
-    def __init__(self, *args, relation, to_field=None, **kwargs):
-        kwargs.setdefault("max_length", 254)
+class LooseRelationField(models.ForeignKey):
+    """A relation that points to one part of the composite key.
+
+    This only points to the first field of the relation, and the second field
+    (e.g. volgnummer/beginGeldigheid) is determined at runtime using query filtering.
+    Without such filter, traversing the relation will produce multiple results.
+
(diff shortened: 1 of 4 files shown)


src/schematools/contrib/django/models.py
```diff
@@ -549,26 +549,28 @@ def __str__(self):
...
 
 
-class LooseRelationField(models.CharField):
-    def __init__(self, *args, relation, to_field=None, **kwargs):
-        kwargs.setdefault("max_length", 254)
-        super().__init__(*args, **kwargs)
-        self.relation = relation
-        self.to_field = to_field
+class LooseRelationField(models.ForeignKey):
+    """A relation that points to one part of the composite key.
```
lbam, line 553: :+1: for docstring

src/schematools/contrib/django/models.py
```diff
@@ -549,26 +549,28 @@ def __str__(self):
...
+    Without such filter, traversing the relation will produce multiple results.
+
+    This setup is typically used for temporal relationships, where the foreign table uses compound
+    keys as identifier. This field type references the first field of that compound relationship.
+
+    At construction, the "to_field" will be set pointing to the correct identifier field.
+    """
+
+    # Disable the unique check made against the target field.
+    # When it's part of a composite key, this can't be unique by itself.
```
lbam, line 566:
```suggestion
    # When it's part of a composite key, this can't always be unique by itself.
```

(It can be unique, it just doesn't have to be.)

lbam (approved):

PR #312, opened 2022-08-04
Fix importing for fields with shortname and no data

roel-devries:
In case of a nm relation the data doesn't need to be there, but in an other table. Don't try to rename if the data is not there.


jjmurre (approved):
LGTM

PR #319, opened 2022-08-15
Style fixes to types.py and less strict flake8 checking

lbam:
Number of mypy errors down to 62, from 66.


jjmurre (approved):
LGTM, good decision to remove some of the very strict checks. We are not on this world to please flake8 checkers :-)

PR #320, opened 2022-08-18
Allow enable_db with an endpoint_url

lbam:
The reason to allow this combination is to generate OpenAPI docs for the remotes. We generate those with drf_spectacular, which looks at the database. Generating OpenAPI docs from the schemas directly would be possible, but only by duplicating all of drf_spectacular's functionality.

For [AB#50780](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/50780).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = amsterdam-schema-tools
-version = 4.1.3
+version = 4.2.0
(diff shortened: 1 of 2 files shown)


jjmurre (approved):
LGTM.

PR #321, opened 2022-08-23
Add validator for the types of identifier fields

lbam:
For [AB#53421](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/53421).

I had to change an unrelated map expression to a comprehension to make flake8 happy.

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -39,7 +39,7 @@ def test_camelcase() -> None:
 
 
 def test_id_auth(here: Path) -> None:
-    dataset = dataset_schema_from_path(here / "files/id_auth.json")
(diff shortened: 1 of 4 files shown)


tests/test_validation.py
```diff
@@ -51,6 +51,20 @@ def test_id_auth(here: Path) -> None:
     assert list(errors) == []
 
 
+def test_id_type(here: Path) -> None:
+    dataset = dataset_schema_from_path(here / "files" / "id_type.json")
```
jjmurre, line 55: I cannot find this `id_type.json`, is it check in into the repo?

lbam: Of course it wasn't, but it is now.

PR #324, opened 2022-09-15
Fix non-loose relations in DSO-API by declaring them loose

lbam:
This is a terrible hack, but at least relational queries such as

    https://api.data.amsterdam.nl/v1/loopfietsnetwerk/edges/?startNode.nodeId=1

should work again.

For [AB#55902](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/55902).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = amsterdam-schema-tools
-version = 4.2.2
+version = 4.2.3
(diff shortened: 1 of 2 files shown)


vdboor:
err. tricky. This changes a lot of other model fields because `is_loose_relation` is used in model_factory()

PR #326, opened 2022-09-29
Use list instead of dict for dataset collections

lbam:
We were passing around dicts keyed on the dataset id. But the dataset id is contained in a DatasetSchema, so we can just use lists instead.

Also, check for inconsistencies in index.json in _schemas_from_url.


vdboor:
Hoe zit het met DSO-API, gebruikt die nog een API hiervan?

PR #327, opened 2022-10-11
Converter to data.overheid.nl CKAN format

lbam:
For AB#41632. Associated CLI command is being written.


src/schematools/ckan/_convert.py
```diff
@@ -0,0 +1,76 @@
+from collections import defaultdict
+from time import strftime
+from typing import Any, Optional
+
+from ..types import DatasetSchema
+
+
+def from_dataset(ds: DatasetSchema) -> dict[str, Any]:
```
vdboor, line 8: I'd use a different name. e.g. `dataset_to_ckan()`? That "from_..*" notation works well as classmethod but now that name exists outside such scope

src/schematools/ckan/_convert.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+# JSON is our primary file type. Not sure if we can submit multiple types for a dataset.
+# Keep this consistent with the mimetype.
+# From https://waardelijsten.dcat-ap-donl.nl/mdr_filetype_nal.json
+JSON = "http://publications.europa.eu/resource/authority/file-type/JSON"
+
+# From https://waardelijsten.dcat-ap-donl.nl/overheid_license.json
+LICENSES: dict[Optional[str], str] = defaultdict(
+    lambda: "http://standaarden.overheid.nl/owms/terms/licentieonbekend",
```
vdboor, line 63: `LICENSES.get("...", UNKNOWN_LICENSE)` would be equally clear to understand.

src/schematools/ckan/_convert.py
lbam: The intention is to `import ckan` instead of `from ckan import from_dataset`.

src/schematools/ckan/_convert.py
lbam: Done.

src/schematools/ckan/_convert.py
vdboor: ok, perhaps ask a few others whether they agree. This will be a public API after all.

vdboor (approved):

src/schematools/ckan/_convert.py
lbam: We hebben net gestemd en het is drie tegen twee vóór deze optie, met de opmerking dat je altijd nog

    from schematools.ckan import from_dataset as dataset_to_ckan

kunt doen ;)

PR #332, opened 2022-10-19
Massive change to id/name/short_name/db_name to fix shortnames

vdboor:
The table/field.name attributes returned shortnames while that should only be used for the database field names. This created a sideeffect where shortnames were used as Django model names, API serializer field names and various other subtle comparison checks.

This change fixes that, and also corrects many incorrect usages of `id` vs `name` vs `db_name`. Each schema object now has attributes like `id` (unaltered), `name` (camel cased), `db_name` (uses shortname) and `python_name` (snake cased).

As these changes broke the lengthy importer logic in various subtle places, which nearly became unfixable without a significant cleanup. By switching from in-place editing to building a new row dict, the magic `Row` class could be replaced with a simple `Provenance` object per value.

Having an improved importer also allowed removing the ugly logic of `get_subfields(add_prefixes=True)` because db_name addresses that. Hence, the difference between `table.fields` vs `table.get_fields()` is also fixed, so it interates over the actual schema fields only. To retrieve inlined-subfields, `table.get_fields(include_subfields=True)` should be used.

Additional properties were added to `schematools.types` so these make it easier to read metadata from the schema fields. For example, many calls to `get_field_by_id()` are no longer needed, since `table.identifier_fields` replaces logic like: `table.get_field_by_id(table.identifier[0])`. The same is applied to many other places where fields were requested by ID.

DatasetSchema:
- python_name (formats as ClassName)
- db_name (formats in snake_case)

DatasetTableSchema:
- python_name (formats as ClassName)
- short_name
- through_fields (for through tables)
- temporal.identifier_field
- temporal.dimensions returns actual fields
- main_geometry_field
- identifier_fields
- db_name() => db_name   - became a property for the typical common usage
- db_name_variant()      - provides the versioned-table support

DatasetFieldSchema:
- python_name
- db_name() => db_name  - became a property for consistency
- is_identifier_part
- is_subfield
- is_temporal => is_temporal_range
- get_dimension_fieldnames() => get_dimension_fields()
- get_subfields() => subfields   - no longer needs prefixes.
- related_fields
- nested_table
- through_table

Deleted obsolete functions:
- get_through_tables_by_id()
- get_fields_by_id()
- _get_fk_fields()


src/schematools/events/consumer.py
```diff
@@ -43,9 +43,8 @@ def _fetch_consumer_params() -> dict:
 
 def consume_events(
     dataset_schemas: list[DatasetSchema],
-    srid: str,
     connection: Connection,
-    topics: tuple(str),
+    topics: list[str],
```
lbam, line 47:
```suggestion
    topics: Iterable[str],
```

?

src/schematools/factories.py
```diff
@@ -1,12 +1,19 @@
-"""Module to hold factories."""
+"""Module for SQLAlchemy-based database table creation."""
```
lbam, line 1: Maybe rename the module while we're here? `tablefactory`?

src/schematools/types.py
```diff
@@ -740,8 +734,16 @@ def __repr__(self):
...
-    def name(self) -> str:
-        return self.get("shortname", self.id)
+    def class_name(self) -> str:
+        """The 'id', but camel cased like a class name."""
+        return toCamelCase(self.id).title()
+
+    @property
+    def shortname(self) -> str:
+        """The shorter name if present, otherwise the ID.
+        Note this is only used to generate human-readable database table names.
```
lbam, line 744:
```suggestion
        This is only used to generate human-readable database table names.
```

src/schematools/types.py
```diff
@@ -1031,27 +1055,25 @@ def db_name(
...
             dataset_prefix = to_snake_case(self.dataset.id)
+
+        shortname = to_snake_case(self.shortname)
         if self.nested_table or self.through_table:
             # We don't automatically shorten user defined table names. Automatically generated
             # names, however, should be shortened as the user has no direct control over them.
-            db_table_name = "_".join(filter(None, (dataset_prefix, to_snake_case(self.name))))
-            additional_underscores = len(list(filter(None, (version_postfix, postfix))))
+            db_table_name = _name_join(dataset_prefix, shortname)
+            additional_underscores = _name_join_count(dataset_prefix, shortname)
```
lbam, line 1064: This is incorrect:

```python
>>> _name_join("foo", "bar")
'foo_bar'
>>> _name_join_count("foo", "bar")
2
```

... but there's only one underscore. (Not introduced by this PR.)

src/schematools/types.py
```diff
@@ -1095,6 +1109,16 @@ def from_dict(cls: type[DTS], obj: Json) -> DTS:  # noqa: D102
...
 
 
+def _name_join(*parts):
+    """Combine items with underscores, and skip empty values."""
+    return "_".join(filter(None, parts))
+
+
+def _name_join_count(*parts):
+    """Tell how many parts are Combine items with underscores, and skip empty values."""
+    return len(list(filter(None, parts)))
```
lbam, line 1119:
```suggestion
    """Counts the number of underscores that _name_join inserts."""
    return sum(1 for part in parts if part) - 1
```

src/schematools/types.py
```diff
@@ -279,6 +278,16 @@ def __hash__(self) -> int:
     def id(self) -> str:  # noqa: A003
         return cast(str, self["id"])
 
+    @cached_property
+    def class_name(self) -> str:
```
lbam, line 282: Why not call this one `python_name` too?

src/schematools/types.py
```diff
@@ -740,8 +734,16 @@ def __repr__(self):
         return f"<{self.__class__.__name__}: {prefix}{self['id']}>"
 
     @property
-    def name(self) -> str:
-        return self.get("shortname", self.id)
+    def class_name(self) -> str:
```
lbam, line 737: Same here, `python_name`?

src/schematools/types.py
```diff
@@ -926,7 +925,7 @@ def is_autoincrement(self) -> bool:
         """Return bool indicating autoincrement behaviour of the table identifier."""
         if self.has_composite_key:
             return False
-        return self.get_field_by_id(first(self.identifier)).is_autoincrement
+        return self.get_field_by_id(self.identifier[0]).is_autoincrement
```
jjmurre, line 944: Killing Guido's darling ;-)

src/schematools/types.py
vdboor: I just copied it. And you'd be surprised:

```
>>> timeit('len(list(filter(None, parts)))', setup='parts=["abc", "def", ""]')
0.3183160940000036
>>> timeit('sum(1 for part in parts if part) - 1', setup='parts=["abc", "def", ""]')
0.40874755899999116
```

src/schematools/events/consumer.py
vdboor: ok, fixed!

src/schematools/types.py
vdboor: done!

src/schematools/factories.py
lbam: Fair enough.

src/schematools/types.py
lbam: Those are not equivalent due to the `- 1` in the `sum` version. The number of underscores is one less than the number of parts joined. That's the real point here, the `sum` stuff was just me getting carried away.

vdboor: ahh I see. yes this code is weird in so many ways. the `- 1` should not be there, as the shortened version is combined with another underscore later on. This is really an attempt to cleanup some hard to read code.

lbam: But then it doesn't match the docstring anymore :)

lbam (approved):

jjmurre (approved):
Looks great! I really like the cleanup up API and especially the summary of it in the commit message. Maybe move this summary to a more prominent place. Esp. for new developers?
I focussed on testing the ndjson import with real BenK data. After a small fix, this now works really well. Let's see how it behaves inside Airflow on ACC.

src/schematools/contrib/django/factories.py
```diff
@@ -262,8 +264,10 @@ def handle_basic(
         if not field.is_primary and field.nm_relation is None:
             # Primary can not be Null
             kwargs["null"] = not field.required
+        if field.has_shortname:
+            kwargs["db_column"] = field.db_name
         if self.value_getter:
-            kwargs = {**kwargs, **self.value_getter(dataset, field)}
+            kwargs.update(self.value_getter(dataset, field))
```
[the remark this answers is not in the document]
lbam: It's `**kwargs`, so it's never shared with anything.

PR #333, opened 2022-10-20
Use only .items() of dict[str, DatasetSchema]

lbam:
Preparation for [AB#59480](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/59480), where I want to change the meaning of the key from id to path.


jjmurre (approved):
LGTM

PR #335, opened 2022-10-24
Skip tests that require the database if not configured

lbam:
This way, pytest will run the tests that it can without a database, so we can set up continuous testing (#331).

The implementation is very ugly, but I couldn't boil it down further. The pytest-django plugin doesn't have good support for skipping, so we need to stop collecting the tests instead of skipping them. This gives misleading output, in the sense that the skipped tests aren't even reported as such.


jjmurre (approved):
LGTM. This is a good as it gets with the `pytest-django` plugin I guess.

PR #337, opened 2022-10-24
Turn many often-used schema fields into cached property.

vdboor:
No need to read the source dict that many times


lbam:
Got any figures on the performance improvement?

lbam (approved):

PR #338, opened 2022-10-25
AB#42523 expand mocker

jjmurre:
Expand the initial implementation of the mocker and refine the faker implementation.

The mocker is implemented as a set of two Django management commands. It can be run on a local dev. environment, it can also be used from a pipeline (Azure or otherwise) that generate mock data.

The `create_mock_data` command generates mock data for all the datasets. It is also possible to limit the scope to a smaller subset using `limit-to` or `skip` cli options. The readme had more information on the possible options.

The `create_mock_data` command does not generate the relations. So, foreign keys are filled with `null` values.

There is a second command `relate_mock_data` that generates the foreign key fields and the through tables for the relations.
The `relate_mock_data` also has some flags to limit its scope (see readme).

The mocking is based on the `factoryboy` and `faker` python modules. For mocking, a separate factory function is created
that is responsible for generating the Django models used in the mocking process. 
Instead of using the regular Django field classes to create fields for the model, there is a factory function that creates the appropriate mock class. This function lives in the package `schematools.contrib.djanog.faker` and takes the following into account when creating a fake field class:

- the type of the field (from amsterdam schema)
- when a field has an `enum`, values from the enum are used for mocking
- min and max configuration in the amsterdam schema definition is taken into account
- if a field is nullable (not `required`) `null` values are sprinkled into the mocked values
- a field can have a `faker` definition in the amsterdam schema, this is string identifier, that needs to be available in the list of declarations in the `faker` package

In the relate step the foreign keys are filled with ids that are pointing to the correct records of the relation. M2M relations are created in the same way by updating the foreign keys in the through tables.
When a relation is pointing to a so-called temporal table, additional columns are updated in the source table (usually, these are the `_indentificatie` and `_volgnummer` postfixed columns.


README.md
```diff
@@ -113,3 +113,28 @@ schema-tools, make sure one of the commit increments the version number in
...
+## Mocking data
+
+The schematools library contains two Django management commands to generate
+mock data. The first one is `create_mock_data` which generates mock data for all
+the datasets that are found at the configured schema location `SCHEMA_URL`.
+
+The `create_mock_data` command has some options, e.g. to change
+the default number of generated records (`--size`), to skip datasets (`--skip`)
+or to use a subset of the datasets (`--limit-to`). To avoid duplicate pk's
+on subsequent runs the `--start-at` options can be used.
```
lbam, line 127: I'm not sure why we need two options to control the datasets. Why not have

    schema create_mock_data

generate mock data for all datasets, and have

    schema create_mock_data foo bar

generate data only for `foo` and `bar`? Right now, I can't even figure out how to use `--limit-to`. Is it

    schema create_mock_data --limit-to="foo bar"
    schema create_mock_data --limit-to=foo,bar
    schema create_mock_data --limit-to=foo --limit-to=bar

?

When the datasets are ordinary cmdline args, `--skip` can just be a boolean flag and inconsistent combinations like `--limit-to=foo --skip=foo` become impossible.

tests/django/test_mocking.py
```diff
@@ -0,0 +1,230 @@
...
+from schematools.contrib.django.faker.relate import relate_datasets
+from schematools.utils import to_snake_case
+
+from .conftest import DATABASE_URL
+
+
+@pytest.mark.django_db
+@pytest.mark.skipif(DATABASE_URL is None, reason="DATABASE_URL not set")
+def test_mocking_creates_data(gebieden_schema, gebieden_dataset):
+    """Prove that mocking create data."""
```
lbam, line 20:
```suggestion
    """Prove that mocking creates data."""
```

Maybe we should get rid of mandatory docstrings, because this one just repeats the function name.

tests/django/test_mocking.py
```diff
@@ -0,0 +1,230 @@
...
+
+    table_schemas = {
+        to_snake_case(t.id): t for t in gebieden_schema.get_tables(include_through=True)
+    }
+    models = {}
+    for cls in schema_models_factory(gebieden_dataset, base_app_name="dso_api.dynamic_api"):
+        models[cls._meta.model_name] = cls
+
+    for model in models.values():
+        assert model.objects.count() == 3
```
lbam, line 32:
```suggestion
    size = 3
    
    create_tables(gebieden_dataset)
    create_data_for(gebieden_dataset, size=size)

    table_schemas = {
        to_snake_case(t.id): t for t in gebieden_schema.get_tables(include_through=True)
    }
    models = {}
    for cls in schema_models_factory(gebieden_dataset, base_app_name="dso_api.dynamic_api"):
        models[cls._meta.model_name] = cls

    for model in models.values():
        assert model.objects.count() == size
```

tests/django/test_mocking.py
```diff
@@ -0,0 +1,230 @@
...
+    create_tables(verblijfsobjecten_dataset)
+    create_tables(afvalwegingen_dataset)
+    create_data_for(afvalwegingen_dataset, size=30)
+
+    models = {}
+    for cls in schema_models_factory(afvalwegingen_dataset, base_app_name="dso_api.dynamic_api"):
+        models[cls._meta.model_name] = cls
+
+    # The containertypes.volume field has a `maximum` of 20 and `minimum` of 5
+    assert models["containertypes"].objects.filter(volume__gt=12, volume__lt=5).count() == 0
```
lbam, line 54: 20 or 12?

tests/django/test_mocking.py
jjmurre: I'm all for it to disable this in flake8. Every developer knows when it makes sense to add a docstring.

README.md
jjmurre: Good point, that's a better cli design. However, `--skip` should not be a boolean, but one or more dataset ids. So that one can create mocks for all dataset but exclude just a few (usually the proxied datasets, because these datasets do not need database tables).  Thinking about it, we have the flag `enable_db` in `datasets_dataset`, so it is even more convenient to always exclude these datasets. I'll implement that.

jjmurre: The reason for introducing `--limit-to` was that the positional arguments can be one or more paths to dataset schema files. However, it would be easy to find out if an argument is a local path or just a dataset identifier.

lbam: If the datasets are positional args, then we can say

    schema create_mock_data --skip foo bar baz

to process everything but foo, bar and baz.

And don't we already have code to handle `--schema-url` specially for local paths for other commands?

jjmurre: The code that uses `--schema-url` is not being used in the Django based mgm. commands (like `import_schemas`). I now applied the same behaviour that `import_schemas` has to these new mgm. commands.
I'll have a look if the `--schema-url` based approach can also be applied to the Django mgm. commands.

jjmurre: It was not too hard to integrate the same schemaloader for the Django mgm. commands, so I added it.

tests/django/test_mocking.py
jjmurre: I added D101 and D102 to the flake8 ignores.

README.md
jjmurre: And also added this to the `import_schemas` Django mgm. command.

tests/django/test_mocking.py
jjmurre: Comment is not very useful anyway, I'll remove it.

README.md
jjmurre: Exclude datasets with an `enable_db` is False.

lbam (approved):

PR #340, opened 2022-10-27
Cleanup

lbam:
Triggered by a test failure that said

    NoSuchModuleError: Can't load plugin: sqlalchemy.dialects:postgres

The [solution](https://stackoverflow.com/a/64698899) was to change my `DATABASE_URL`, after which I went `git grep`'ing for `postgres://` and got two hits.


jjmurre (approved):
LGTM

PR #341, opened 2022-10-27
Fix circular import

lbam:
Importing types first and utils second works, but the other way around gave an exception. That's gone now.

TODO: refactor further by splitting utils into a loading module and a naming (snakes/camels) module.

Removed an import of loaders in utils that didn't seem to do anything.


vdboor (approved):

lbam:
Doesn't work as intended: pytest crashes.

PR #342, opened 2022-10-27
Split naming module off of utils

lbam:
Triggered by circular import exceptions that occur when utils is imported before typing (but not the other way around). Having a utils module is a code smell anyway.

This replaces #341.


jjmurre (commented):
I assume that DSO API and the airflow repo also needs to change imports of `to_snake_case` and `toCamelCase`?

PR #343, opened 2022-10-31
AB#60440 (datasets on subpaths)

jjmurre:
When a dataset is in a subfolder (e.g. haalcentraalbag), a configuration with a SCHEMA_URL pointing to the local filesystem, does not find these datasets.

This PR fixes that problem.

### DIFF
--- a/src/schematools/utils.py
+++ b/src/schematools/utils.py
@@ -188,6 +188,20 @@ def dataset_schema_from_path(
     return types.DatasetSchema.from_dict(ds)
 
 
+def _get_relative_dataset_path(schemas_path: Path | str, dataset_id: str) -> Path:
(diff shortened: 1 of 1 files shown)


src/schematools/utils.py
```diff
@@ -188,6 +188,20 @@ def dataset_schema_from_path(
     return types.DatasetSchema.from_dict(ds)
 
 
+def _get_relative_dataset_path(schemas_path: Path | str, dataset_id: str) -> Path:
+    """Gets the path of a dataset relative to `dataset_path`."""
```
DonQueso89, line 192:
```suggestion
    """Gets the path of a dataset relative to `schemas_path`."""
```

DonQueso89 (dismissed):

DonQueso89 (approved):

PR #351, opened 2022-11-08
Always fill in "notes" field in CKAN format

lbam:
data.overheid.nl requires a non-empty string, or it will not accept our output. "parkeervakken" has no description and no title, so just put the id in the notes field.


jjmurre (approved):
LGTM

PR #354, opened 2022-11-16
non Amsterdam schema field for ingestion guidance

cvriel:
The GOB data processing in Airflow utilizes the NDJSONImporter class to ingest .ndjson files. The processing logic relies on a non Amsterdam schema defined field called: "cursor". However, during the read of the .ndjson file, only schema defined fields are currently taken into account. Leading to an error when the processing logic wants to retrieve the value of "cursor".

Context:
The source field "cursor" is used to track the position of the next data row for the next batch. This "cursor" field is merely used as a guidance during ingestion and therefore does not to be present in the target table.

Changes:
- Class method `parse_records` provides an optional argument for extra fields that could be needed during ingestion (but not to be present in the target table).
- Class method `load_file` provides an optional argument for extra fields that could be needed during ingestion (but not to be present in the target table).

### DIFF
--- a/src/schematools/importer/base.py
+++ b/src/schematools/importer/base.py
@@ -267,13 +267,29 @@ class BaseImporter:
                 db_schema_name,
             )
 
-    def load_file(self, file_name: Path, batch_size: int = 100, **kwargs: Any) -> dict | None:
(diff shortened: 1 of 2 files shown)


vdboor:
Solution implemented via #360

PR #356, opened 2022-11-17
Introduce new loader, and remove singleton _DatasetCollection class

vdboor:
This loader replaces the logic in `schematools.utils`. Instead of having separate functions, there is a single interface that can be implemented for every type of loading (URL, filesystem or database or from cache).

This interface also allows retrieval of other table versions (`loader.get_table()`), which is the main motivation for this change.

Secondary benefit is that the `_DatasetCollection` could be removed. Instead of having a global cache, the datasets are only cached within the same loader instance (or `QuerySet`) that retrieves the data.

However, this will break code that assume datasets are globally cached. This affected both the Django management commands, and unit tests. The tests use a shared `schema_loader` fixture instead, so all schemas from the same test are cached.


src/schematools/loaders.py
```diff
@@ -1,33 +1,44 @@
...
 
     def get_dataset(self, dataset_id: str, prefetch_related: bool = False) -> DatasetSchema:
         """Gets a dataset for dataset_id."""
         raise NotImplementedError
 
+    def get_table(self, dataset: DatasetSchema, table_ref) -> DatasetTableSchema:
+        """Retrieves a versioned table by reference"""
+        raise NotImplementedError
+
+    def get_dataset_path(self, dataset_id) -> str:
```
lbam, line 38:
```suggestion
    def get_table(self, dataset: DatasetSchema, table_ref: str) -> DatasetTableSchema:
        """Retrieves a versioned table by reference"""
        raise NotImplementedError

    def get_dataset_path(self, dataset_id: str) -> str:
```

src/schematools/loaders.py
```diff
@@ -36,45 +47,458 @@ def get_all_datasets(self) -> dict[str, DatasetSchema]:
...
+        return f"{self.__class__.__name__}({self.schema_url!r})"
+
+    @classmethod
+    def get_root(cls, dataset_file: Path | str) -> Path:
+        """Resolve the real root folder.
+        This makes sure that the datasets-path value will be correct,
+        even when a subfolder is selected for importing files.
+        """
+        if isinstance(dataset_file, str):
+            dataset_file = Path(dataset_file)
```
lbam, line 277: `Path` is idempotent (`Path(Path("foo")) == Path("foo")`), so we don't need to check:

```suggestion
        dataset_file = Path(dataset_file)
```

src/schematools/loaders.py
```diff
@@ -36,45 +47,458 @@ def get_all_datasets(self) -> dict[str, DatasetSchema]:
...
+        (e.g. useful for unit testing). It will however not be possible to resolve relations
+        to other datasets when those datasets use same free-format for naming their files.
+
+        Concluding, any relations will only resolve when:
+
+        * the dataset follows the ``name/dataset.json`` convention;
+        * or when the related dataset is also read and cached by the same loader instance.
+        """
+        if isinstance(dataset_file, str):
+            dataset_file = Path(dataset_file)
```
lbam, line 309:
```suggestion
        dataset_file = Path(dataset_file)
```

lbam (dismissed):

lbam (approved):

PR #360, opened 2022-11-17
Bundle 'source' data to imported records for debugging and field tracking

vdboor:
This allows projects like dataservices-airflow to track the "cursor" field in the BenK/GOB data import. That field is not part of the schema, but needed to retrieve more data. As a side effect, the typing in code became much easier to read, and debugging shows source data too.

### DIFF
--- a/src/schematools/importer/base.py
+++ b/src/schematools/importer/base.py
@@ -95,6 +95,25 @@ class Provenance:
             return source[self.provenance]
 
 
+class Record(dict):
+    """A single generated row from the importer.
+    Each key matches the database field name, value is the data that should be written.
+
+    This is a tiny wrapper around dict for simplicity, and allowing the source record
+    to be accessed. The source record is useful debugging information,
+    and some implementations need to retrieve non-schema fields
+    (such as a "cursor" field in the last record).
+    """
+
+    def __init__(self, data, source: dict):
+        super().__init__(data)
+        self.source = source
+
+    def __repr__(self):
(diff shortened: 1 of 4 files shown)


jjmurre (approved):
LGTM, Nice and clean solution!

src/schematools/importer/base.py
```diff
@@ -95,6 +95,25 @@ def resolve(self, source: dict) -> Any:
             return source[self.provenance]
 
 
+class Record(dict):
+    """A single generated row from the importer.
+    Each key matches the database field name, value is the data that should be written.
+
+    This is a tiny wrapper around dict for simplicity, and allowing the source record
```
lbam, line 102: ", and allowing" => "that allows"

src/schematools/importer/base.py
```diff
@@ -95,6 +95,25 @@ def resolve(self, source: dict) -> Any:
             return source[self.provenance]
 
 
+class Record(dict):
+    """A single generated row from the importer.
+    Each key matches the database field name, value is the data that should be written.
```
lbam, line 100: There are no `key` and `value` fields. I suppose those were merged into `data`?

src/schematools/importer/base.py
vdboor: good one! included in #361

src/schematools/importer/base.py
vdboor: I've reworded it in #361 - it's about each dict key and value.

PR #363, opened 2022-11-24
Fix importer to handle snake-cased table ID's in other cases too.

vdboor:
tested with 'dataverwerkingen' that this fixes their airflow jobs.

### DIFF
--- a/src/schematools/factories.py
+++ b/src/schematools/factories.py
@@ -61,6 +61,10 @@ def tables_factory(
     metadata = metadata or MetaData()
     dataset_tables = dataset.get_tables(include_nested=True, include_through=True)
 
+    # For compatibility and consistency, allow snake-cased table ID's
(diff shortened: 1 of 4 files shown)


src/schematools/factories.py
```diff
@@ -61,6 +61,10 @@ def tables_factory(
     metadata = metadata or MetaData()
     dataset_tables = dataset.get_tables(include_nested=True, include_through=True)
 
+    # For compatibility and consistency, allow snake-cased table ID's
+    db_table_names = _snake_keys(db_table_names)
+    db_schema_names = _snake_keys(db_schema_names)
```
lbam, line 66: If we do

```suggestion
    db_table_names = _snake_keys(db_table_names or {})
    db_schema_names = _snake_keys(db_schema_names or {})
```

we don't need the `if` inside `_snake_keys`.

vdboor: done!

lbam (approved):

PR #365, opened 2022-12-01
Fix importer to skip M2M table data when limit_tables_to is used.

vdboor:
The previous importer logic (from 6951758398444d989399e2f16272d1e56b7d1325) implicitly did this by iterating over the insert_statements, and therefore silently dropping records that didn't have a corresponding table.

This new logic can detect the issue and warn for it so invalid importer logic won't silently drop records.


jjmurre (approved):
LGTM, nice improvement for users of the library.

PR #366, opened 2022-12-01
Cleanups for geometry index creation

vdboor:
When working on geo index code, I found that column creation happened across various files (`schematools.importer` and `schematools.factories`). This moves all SQLAlchemy factory logic in `schematools.factories` so new developers can find the complete code at once. This also fixes a subtle SRID bug in the old code.

* Centralize 3D coordinate system detection (wasn't done everywhere the same)
* Fix SRID type when creating columns with SQLAlchemy
* Fix inconsistent geometry index name (it was created after all, but very hidden in geoalchemy2)


src/schematools/__init__.py
```diff
@@ -1,11 +1,28 @@
...
 MAX_TABLE_NAME_LENGTH: Final[int] = 63  # limitation of PostgreSQL
 TMP_TABLE_POSTFIX: Final[str] = "_new"
 TABLE_INDEX_POSTFIX: Final[str] = "_idx"
 DATABASE_SCHEMA_NAME_DEFAULT: Final[str] = "public"
+
+# Defaults
 DEFAULT_SCHEMA_URL: Final[str] = "https://schemas.data.amsterdam.nl/datasets/"
 DEFAULT_PROFILE_URL: Final[str] = "https://schemas.data.amsterdam.nl/profiles/"
+
+# Common coordiate reference systems
```
lbam, line 16:
```suggestion
# Common coordinate reference systems
```

lbam (dismissed):

lbam (approved):

PR #370, opened 2022-12-05
Fix relative path import for schema files

vdboor:
This fixes cases like: `./manage.py import_schemas ../../amsterdam-schema/datasets/bag/dataset.json`


jjmurre (approved):
LGTM

PR #380, opened 2022-12-13
Fix processing `id` field twice

roel-devries:
When the data contains an `id` during import, its processed twice. Its safe to skip the second iteration.

### DIFF
--- a/src/schematools/importer/ndjson.py
+++ b/src/schematools/importer/ndjson.py
@@ -92,7 +92,7 @@ class TableFieldMapper:
         for field in self.dataset_table.get_fields(include_subfields=True):
             if field.id == "schema" and field.type.startswith("https://"):
                 continue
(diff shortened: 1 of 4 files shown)


src/schematools/importer/ndjson.py
```diff
@@ -92,6 +92,10 @@ def parse_object(self, source: dict) -> dict[str, list[Record]]:
         for field in self.dataset_table.get_fields(include_subfields=True):
             if field.id == "schema" and field.type.startswith("https://"):
                 continue
+            if field.id == "id" and self.dataset_table.has_composite_key:
```
lbam, line 95: Needs to use the same condition that `_fill_composite_pk` uses:

```suggestion
            if field.id == "id" and self.dataset_table.is_autoincrement and self.dataset_table.has_composite_key:
```

lbam (approved):

PR #393, opened 2022-12-15
Add sqlmigrate_schema management command to generate migration SQL

vdboor:
This builds a ProjectState for the old and new table, allowing the migration engine to generate the proper set of operations.


lbam:
What's with [src/schematools/contrib/django/management/commands/validate_schemas.py](https://github.com/Amsterdam/schema-tools/pull/393/files#diff-d831742634f0079df0b9fbffccb24c9bffc0ca5b265c1dd78299bf8875ae2b69)?

vdboor:
@lbam that was an empty file. Maybe I should move that to a separate commit.

lbam (approved):

PR #394, opened 2022-12-16
Remove unused property DatasetSchema.identifier

lbam:
This method gives access to the "identifier" property on datasets. While some datasets have that, it is not in line with the Schema spec and never used. When I replace the property definition by

    @property
    def identifier(self) -> str:
        raise ValueError("crash")

the tests for schema-tools and DSO-API all still run.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -323,11 +323,6 @@ class DatasetSchema(SchemaType):
         return self.get("license")
 
     @property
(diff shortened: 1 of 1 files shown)


jjmurre (approved):
Yes, can be removed. Originally, this was used as part of the temporal definitions for a dataset. However, temporal info has been moved 'down' to the table definitions.

PR #399, opened 2022-12-19
Loading and validating publishers under metaschema 1.3.0 and 2.0.0

DonQueso89:
* Support for loading and validating publishers from the schema server as top-level objects
* Support for inlining  and validating referenced publisher objects inside datasets as defined in metaschema 2

The main idea is that the publishers location is a well known directory which can be derived from SCHEMA_URL (i.e. a /publishers directory at the same level as /datasets).

I will release a new schema-tools version in a separate PR because we need to deploy metaschema 2 with the properly structured publishers directory first. 

All commands have been tested against a local amsterdam-schema devserver

Also, all test fixtures had to be updated because (ofcourse) they were not compliant with the metaschema

### DIFF
--- a/tests/files/datasets/brk.json
+++ b/tests/files/datasets/brk.json
@@ -14,9 +14,7 @@
       "version": "1.0.0",
       "temporal": {
         "identifier": "volgnummer",
-        "dimensions": {
-          "geldigOp": ["beginGeldigheid", "eindGeldigheid"]
-        }
+        "dimensions": { "geldigOp": ["beginGeldigheid", "eindGeldigheid"] }
       },
       "schema": {
         "$id": "https://github.com/Amsterdam/schemas/brk/kadastraleobjecten.json",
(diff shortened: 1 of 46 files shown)


tests/files/datasets/brk.json
```diff
@@ -1088,7 +1021,7 @@
           },
           "naam": {
             "type": "string",
-            "description": "De officiële vastgestelde gemeentenaam."
+            "description": "De offici\u00eble vastgestelde gemeentenaam."
```
Yrob, line 1024: offici\u00eble

tests/files/datasets/gebieden.json
```diff
@@ -45,12 +43,12 @@
...
           "code": {
             "type": "string",
-            "description": "Officiële code van het object."
+            "description": "Offici\u00eble code van het object."
           },
           "beginGeldigheid": {
             "type": "string",
             "format": "date",
-            "description": "De datum waarop het object is gecreëerd."
+            "description": "De datum waarop het object is gecre\u00eberd."
```
Yrob, line 51: Waar komen al die escape chars vandaan? JSON ondersteund unicode

Yrob (dismissed):

tests/files/datasets/gebieden.json
DonQueso89: hmm heb t ff snel met een python script gedaan, zal wel een ascii decoding ergens tussen in perongeluk zijn geweest

Yrob (approved):

PR #400, opened 2022-12-20
Metaschema aware schematools

DonQueso89:
* Schematools now defines which metaschemas it can work with and restricts its usage to these versions
* The validation logic is updated to support validation attempts on multiple metaschema versions so we can transition datasets from one major version to another in amsterdam-schema
* `DatasetTableSchema.validate` and `DatasetTableSchema._resolve` have been removed because they are remnants of an abandoned atempt to make datase-schema an table-schema compliant JSON schemas

### DIFF
--- a/src/schematools/cli.py
+++ b/src/schematools/cli.py
@@ -465,12 +465,10 @@ def validate_publishers(schema_url: str, meta_schema_url: tuple[str]) -> None:
         SCHEMA_URL=https://example.com/datasets, the publishers are extracted from
         https://example.com/publishers.
     """  # noqa: D301,D412,D417
-    ordered = sorted(
-        [(u, version_from_metaschema_url(u)) for u in set(meta_schema_url)],
-        key=lambda t: t[1],
+    for meta_schema_version, url in sorted(
+        [(version_from_metaschema_url(u), u) for u in set(meta_schema_url)],
         reverse=True,
-    )
-    for url, meta_schema_version in ordered:
+    ):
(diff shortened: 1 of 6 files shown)


src/schematools/cli.py
```diff
@@ -375,8 +388,9 @@ def validate(
     \b
         DATASET_ID: id of the dataset.
         META_SCHEMA_URL: URL where the meta schema for Amsterdam Schema definitions can be found.
+        If multiple are given, schematools will try to validate against the largest version,
+        working backwards and stopping at the first version that the objects are valid against.
```
lbam, line 392: This is rather complicated. What is the use case? Can't it be handled by just calling `schema validate` twice?

src/schematools/cli.py
```diff
@@ -427,32 +465,55 @@ def validate_publishers(schema_url: str, meta_schema_url: str) -> None:
...
-            jsonschema.validate(
-                instance=publisher.json_data(),
-                schema=meta_schema,
-                format_checker=draft7_format_checker,
+    ordered = sorted(
+        [(u, version_from_metaschema_url(u)) for u in set(meta_schema_url)],
+        key=lambda t: t[1],
+        reverse=True,
+    )
+    for url, meta_schema_version in ordered:
```
lbam, line 473:
```suggestion
    for meta_schema_version, url in sorted(
        [(version_from_metaschema_url(u), u) for u in set(meta_schema_url)],
        reverse=True,
    ):
```

lbam (approved):

PR #403, opened 2022-12-22
Hack to make CKAN uploader work with data-acc.overheid.nl

lbam:
Apparently, titles are identifiers. Leaves me wondering what the identifier field is for, because its output never shows up anywhere.

Fixes [AB#61176](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/61176).


jjmurre (approved):
LGTM, hope this helps.

PR #409, opened 2023-01-16
Better error messages from schema batch-validate

lbam:
`jsonschema.ValidationError.message` is much less informative than its `__str__`. The former is also squashed into a single line that gets trimmed down by GitHub Actions so that only a part of the schema JSON appears in the UI.


jjmurre (approved):
Nice!

PR #414, opened 2023-01-19
AB#61561 Support object fields in amsterdam schema

jjmurre:
Objects fields that are not a relation are now 'flattened' to fit in the relational database. So, a schema definition with the following field:

```
"naamGebruik": {
  "type": "object",
  "properties": {
      "code": {
	  "type": "string"
      },
      "omschrijving": {
	  "type": "string"
      }
  }
}

```
will generate a Django model that has fields `naam_gebruik_code` and `naam_gebruik_omschrijving`.

Furthermore, a json type field has been added, so the following field:

```
"soortGrootte": {
    "type": "object",
    "format": "json"
}

```
will generate a jsonb field in the Postgres database.

NB. the structure of this json field is not validated! So, the only check is that the postgres database ensures that the data can be parsed as json.

Also the mocker has been updated to correctly supply these object fields with fake data.

The ndjson importer has been updated so that incoming ndjson records with object fields are processed in the correct way.

The relevant tests have been changed and new tests have been added.

### DIFF
--- a/src/schematools/factories.py
+++ b/src/schematools/factories.py
@@ -10,7 +10,7 @@ from typing import DefaultDict, cast
 
 from geoalchemy2.types import Geometry
 from psycopg2 import sql
-from sqlalchemy import BigInteger, Boolean, Date, DateTime, Float, Numeric, String, Time
+from sqlalchemy import JSON, BigInteger, Boolean, Date, DateTime, Float, Numeric, String, Time
 from sqlalchemy.sql.schema import Column, Index, MetaData, Table
 from sqlalchemy.types import ARRAY
 
@@ -32,6 +32,7 @@ FORMAT_MODELS_LOOKUP = {
     "date-time": DateTime,
     "uri": String,
     "email": String,
+    "json": JSON,
 }
 
 JSON_TYPE_TO_PG = {
--- a/src/schematools/importer/ndjson.py
+++ b/src/schematools/importer/ndjson.py
@@ -172,6 +172,8 @@ class TableFieldMapper:
         elif field.id in self.inactive_relation_info:
             # Convert nested object to JSON string
             return json.dumps(value)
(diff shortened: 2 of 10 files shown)


src/schematools/factories.py
```diff
@@ -10,7 +10,7 @@
 
 from geoalchemy2.types import Geometry
 from psycopg2 import sql
-from sqlalchemy import BigInteger, Boolean, Date, DateTime, Float, Numeric, String, Time
```
lbam, line 13: I'm surprised that isort is okay with this, but whatever.

src/schematools/factories.py
```diff
@@ -10,7 +10,7 @@
...
+from sqlalchemy import JSON, BigInteger, Boolean, Date, DateTime, Float, Numeric, String, Time
```
lbam, line 13: I'm surprised isort is okay with this, but whatever.

src/schematools/importer/ndjson.py
```diff
@@ -172,7 +172,7 @@ def _format_db_value(self, field: DatasetFieldSchema, value) -> Any:
         elif field.id in self.inactive_relation_info:
             # Convert nested object to JSON string
             return json.dumps(value)
-        elif isinstance(value, (dict, list)):
+        elif isinstance(value, (dict, list)) and not field.is_json_object:
```
lbam, line 175: This passes lists through, is that intended?

jjmurre: That is intended, however writing it like this is not really clear. I changed it so that the `is_json_object` will become a separete case in the if/elif/else.

src/schematools/factories.py
jjmurre: Yeah, I changed it manually.

lbam (approved):

PR #423, opened 2023-01-30
Add path-validator whitelist for haalcentraal ids

jjmurre:


lbam (approved):

tests/test_validation.py
```diff
@@ -73,10 +73,10 @@ def test_id_matches_path(here: Path, schema_loader) -> None:
     error = next(errors)
     assert error
     assert error.validator_name == "ID does not match file path"
-    assert (
-        """Id of the dataset stadsdelen does not match the parent directory /app/tests/files/datasets/regios."""  # noqa: E501
-        in error.message
+    assert error.message.startswith(
+        "Id of the dataset stadsdelen does not match the parent directory"
     )
+    assert error.message.endswith("files/datasets/regios.")
```
lbam, line 79: Maybe use a regexp?

```suggestion
    assert re.match(r"^Id of the dataset stadsdelen does not match the parent directory .*/files/datasets/regios\.$", error.message)
```

jjmurre: Sounds like a good plan, less verbose an still readable.

PR #424, opened 2023-01-30
Print error path as is from batch-validate

lbam:
Apparently, the path in a jsonschema.ValidationError can contain any types, as seen in the TypeError at

    https://github.com/Amsterdam/amsterdam-schema/actions/runs/4045114336/jobs/6956149861

Just print it in Python syntax. We'll figure it out.

### DIFF
--- a/src/schematools/cli.py
+++ b/src/schematools/cli.py
@@ -578,7 +578,7 @@ def batch_validate(
                 )
             except (jsonschema.ValidationError, jsonschema.SchemaError) as struct_error:
                 errors[schema_file][meta_schema_version].append(
-                    f"{struct_error}: ({', '.join(struct_error.path)})"
+                    f"{struct_error}: {struct_error.path}"
                 )
 
             for sem_error in validation.run(dataset, main_file):


src/schematools/cli.py
```diff
@@ -578,7 +578,7 @@ def batch_validate(
                 )
             except (jsonschema.ValidationError, jsonschema.SchemaError) as struct_error:
                 errors[schema_file][meta_schema_version].append(
-                    f"{struct_error}: ({', '.join(struct_error.path)})"
+                    f"{struct_error}: {struct_error.path}"
```
jjmurre, line 581: Or maybe cast elements to str: `.join(str(e) for e in struct_error.path)`?

DonQueso89 (approved):

src/schematools/cli.py
lbam: I considered that, but putting too long an expression in an f-string makes the code awkward to read. Also, this is unambiguous:

```python
>>> f"{1}" == f"{'1'}"
True
```

Not sure if this can happen, but I don't want to worry about it either :)

PR #428, opened 2023-02-07
AB#70024 object fields for api

jjmurre:


src/schematools/types.py
```diff
@@ -1567,6 +1574,7 @@ def subfields(self) -> list[DatasetFieldSchema]:
 
         If self is not an object or array, the return value is an empty iterator.
         """
+
```
lbam, line 1577: I don't mind this, but do black and flake8 like it? I thought one of them warns about space after docstring.

src/schematools/types.py
```diff
@@ -1265,8 +1267,13 @@ def name(self) -> str:
...
         In general, the "id" field is already camel-cased,
         but in case that didn't happen this property will fix that.
+        For nested subfields (not a relation), the name needs to be prefixed
+        for the external world.
         """
-        return toCamelCase(self._id)
+        name = self._id
+        if self._parent_field is not None and self._parent_field.is_nested_object:
+            name = f"{self._parent_field._id}_{self._id}"
+        return toCamelCase(name)
```
lbam, line 1276: Recursion for deep nesting:

```suggestion
        name = toCamelCase(self._id)
        if self._parent_field is not None and self._parent_field.is_nested_object:
            name = f"{self._parent_field.name}_{name}"  # Note: recursive use of .name.
        return name
```

lbam (approved):

PR #431, opened 2023-02-14
Pin sqlalchemy and pg-grant to be in sync with each other

jjmurre:
And fix problems with the validator that is crashing, because of incompatibility between SA < 1.4 en pg-grant == 0.4.0


lbam:
Replaced by #432.

PR #434, opened 2023-02-21
Fix validation commands

lbam:
The schema validate and schema batch-validate commands no longer worked with local schema files, because they tried to parse a version number out of the "URL" (filename). Rather than complicate this even further, just remove the version check so the commands work again.


DonQueso89:
I dont understand why we are removing this. I thought we had agreed to try to improve the compatibility between the metaschemas and schematools. Why dont we add a function that can resolve the version of the metaschema from a file?

PR #438, opened 2023-02-27
Fix repr of SchemaType in case of missing keys

lbam:
I just got a "maximum recursion depth exceeded" error from a broken schema. The issue is that exceptions use its __repr__, which calls qualified_id, which looks up the 'id', which causes a trap to __missing__, which again calls __repr__. This cycle is now broken.


jjmurre (approved):
👍🏼

PR #444, opened 2023-03-16
Fix code to enable attributes on a relation

jjmurre:
For the BenK datasets relations can have their own "geldigheid". To store this extra information, a through table is needed.

This is already in place for NM relations and for
1-N relations having a temporal table as the target.

However, for 1-N relations to a non-temporal table no through table is being created.

The condition that was in place to determine if a 1-N relation needs a through table was based on the temporality of the target. This condition has been changed now. The new condition is as follows. If the relation does contains fields that are not part of the key a through table will be defined for this relation.

An example:
The table `brk.tenaamstelling` has the following 1-N relation field:

  "isGebaseerdOpStukdeel": {
    "type": "object",
    "properties": {
      "identificatie": {
        "type": "string"
      },
      "beginGeldigheid": {
        "type": "string",
        "format": "date-time"
      }
    },
    "relation": "brk:stukdelen"
  }

The target table `brk:stukdelen` only has a single PK `identificatie`. So, there is an extra non-key field `beginGeldigheid` on the relation. This will define a through table for this 1-N relation.

### DIFF
--- a/src/schematools/types.py
+++ b/src/schematools/types.py
@@ -7,6 +7,7 @@ import json
 import logging
 import re
 import typing
+import warnings
 from collections import UserDict
 from collections.abc import Mapping
 from enum import Enum
@@ -630,7 +631,6 @@ class DatasetSchema(SchemaType):
         # For both types of through tables (M2M and FK), we add extra fields
         # to the table (see docstring).
         if field.is_through_table:
-
             dim_fields = {}
 
             if field.is_object:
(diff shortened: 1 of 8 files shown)


tests/test_ndjson.py
```diff
@@ -467,6 +467,31 @@ def test_ndjson_import_with_shortnames_in_identifier(here, engine, hr_schema, db
     assert records == [{"sbi_act_nr": "12"}, {"sbi_act_nr": "16"}]
 
 
+def test_ndjson_import_with_attributes_on_relation(
+    here, engine, brk_schema, verblijfsobjecten_schema, dbsession
+):
+    """Prove that extra attributes on a relation does not interfer with ndjson import."""
```
Yrob, line 473: """Prove that extra attributes on a relation do not interfere with ndjson import."""

src/schematools/types.py
```diff
@@ -1651,7 +1665,7 @@ def through_table(self) -> DatasetTableSchema | None:
 
     @cached_property
     def is_relation_temporal(self):
```
Yrob, line 1667: Moet deze functie nu niet eigenlijk een andere naam hebben? Maar wat dan?

Yrob (approved):

PR #448, opened 2023-03-22
Pin SQLAlchemy to >=1.4, < 2.0

jjmurre:
Because schematools is used from Airflow and the
current Airflow needs SA >= 1.4

Also bumped geoalchemy2 and unpinned pg-grant.

Bumping SQ to >= 2.0 creates a lot of imcompatibilities, so let's stick with < 2.0 for now.

### DIFF
--- a/src/schematools/events/full.py
+++ b/src/schematools/events/full.py
@@ -5,6 +5,7 @@ import json
 import logging
 from collections import defaultdict
 
+from sqlalchemy import inspect
 from sqlalchemy.engine import Connection
 
 from schematools.events import metadata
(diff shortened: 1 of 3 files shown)


src/schematools/events/full.py
```diff
@@ -243,7 +245,8 @@ def __init__(
             }
             self.geo_fields = defaultdict(lambda: defaultdict(list))
             for table_id, table in tfac.items():
-                if not table.exists():
+                if not inspector.has_table(table.name):
+                    # if not table.exists():
```
DonQueso89, line 249: is dit intended?

DonQueso89 (approved):

src/schematools/events/full.py
jjmurre: Nee, foutje, had weggemoeten.

jjmurre: Heb het weggehaald en gepushed.

PR #449, opened 2023-03-28
Fix condition for 1-N through tables.

jjmurre:
The condition that determines if 1-N relation gets a through tables was change to only depend on the fact if the relation has extra fields.

However, for a 1-N relation to a temporary table *without* extra field, we still need to have a through table, so the code has been changed occordingly.


barrydaniels-nl (approved):
Looks good.

PR #450, opened 2023-03-28
Simplify EventProcessor. Relations will be imported as separate objects. Support for 'object' type.

JKoetsier:
Removed the junction table logic and the DataSplitter. Relations are treated as separate objects from a GOB perspective. We'll publish the relations this way as well. This version of the EventProcessor works well with rel objects.

One side note is that we still need to update the object tables from the rel objects, because the API and perhaps some RefDB consumers are still relying on those fields. Will implement that in a subsequent PR.

We'd rather update the object tables from the rel events than the other way around, as rel events are always generated when the relate process has run; this means that these are always up-to-date. In the old situation the relations were updated with the source object (left side of the relation). However, these events are generated when the source object is updated, but the relate process always runs after this. Meaning that, in the old situation, when relations were updated, the old version was always maintained in the ref DB. By using the rel objects als the source of truth, the ref DB is always up-to-date. Plus it removes a lot of the complexity =)


jjmurre (approved):
LGTM

PR #457, opened 2023-04-06
Add support for full (initial) load

JKoetsier:
The event producer initiates a full load, after which the event consumer/event processor replaces the current table.

How it works:
- EventProducer generates ADD events for all rows in the table. The events have `full_load_sequence` in the header.
- The first event has the `first_of_sequence` header set. The event processor will make sure there is an empty tmp table and writes all events to that table.
- The last event has the `last_of_sequence` header set. The event processor will add this last event to the tmp table and replaces the data in the live table with the data in the tmp table, and removes the tmp table.

With this, we can easily turn off the airflow scripts one by one, and periodically initiate a full load to be sure the data in the ref DB is correct. If, for some reason, the data in the refDB is corrupt, we can always initiate a full load by starting the EventProducer with the 'full load' argument.


jjmurre (approved):
LGTM.

PR #463, opened 2023-04-13
Replace UnlimitedCharField with our own implementation

lbam:
... because django_postgres_unlimited_varchar has no license, so it's unclear what we can use it for.


jjmurre (approved):
LGTM.

PR #473, opened 2023-05-23
Make export to csv/jsonlines less memory hungy

jjmurre:
The SQLAlchemy `execute` is loading a full set of records in memory, leading to 4Gb of mem consumption for large datasettables like brk.kadatastraleobjecten.

The new SA 1.4 that we are using already has a nice streaming/batching mechanisme to reduce this memory consumption.

The SA engine has been changed now to be able to use the new connection. No failing tests, so it seems this change is backwards compatible.

### DIFF
--- a/src/schematools/exports/jsonlines.py
+++ b/src/schematools/exports/jsonlines.py
@@ -55,18 +55,20 @@ class JsonLinesExporter(BaseExporter):  # noqa: D101
         query = select(self._get_columns(sa_table, table))
         if self.size is not None:
             query = query.limit(self.size)
-        for r in self.connection.execute(query):
-            writer.write({toCamelCase(k): row_modifier[k](v) for k, v in dict(r).items()})
+        result = self.connection.execution_options(yield_per=1000).execute(query)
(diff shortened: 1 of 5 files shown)


src/schematools/exports/jsonlines.py
```diff
@@ -55,18 +55,23 @@ def write_rows(  # noqa: D102
...
             query = query.limit(self.size)
-        for r in self.connection.execute(query):
-            writer.write({toCamelCase(k): row_modifier[k](v) for k, v in dict(r).items()})
+        result = self.connection.execution_options(yield_per=1000).execute(query)
+        for partition in result.partitions():
+            for r in partition:
+                writer.write({toCamelCase(k): row_modifier[k](v) for k, v in dict(r).items()})
+
+        # for r in self.connection.execute(query):
+        #     writer.write({toCamelCase(k): row_modifier[k](v) for k, v in dict(r).items()})
```
DonQueso89: same als in csv.py?

DonQueso89 (approved):

PR #476, opened 2023-05-24
Model mocker field naming 

DonQueso89:
The mocker factory logic is passing a DatasetTableSchema object with stripped down relation fields and the camel-cased db_name as identifiers for fields. This goes wrong when an identifier x passed to `to_snake_case(toCamelCase(x))` does not produce x. Which is the case for "eersteHNId". Since capital letters should be recognized as a word boundary, Ive expanded the definition of the camel case word boundary regex to include capital letters in between other capital letters.

This would have fixed the original problem with the mangled DatasetTableSchema passed to model_mocker_factory but it seemed a bit weird to me to implicitly rely on toCamelCase and to_snake_case being invertible. Ive therefore chosen to explicitly mangle the DatasetTableSchema fields with the db_name and document it. 

I think this is not an optimal solution. The best would be to pass an intact DatasetTableSchema (as the test that ive skipped asserts) that matches the original schema to the model mocker and somehow tell the mocker how to properly extract the db_name. This would require me to take some more time and look deeper into mocker and model factories.

### DIFF
--- a/tests/test_naming.py
+++ b/tests/test_naming.py
@@ -18,6 +18,7 @@ def test_toCamelCase() -> None:
     # mind the lower case "i" after "33". It should be made upper case
     assert toCamelCase("numbers33inTheMiddle44") == "numbers33InTheMiddle44"
     assert toCamelCase("per_jaar_per_m2") == "perJaarPerM2"
+    assert toCamelCase("eerste_h_n_id") == "eersteHNId"
 
     with pytest.raises(ValueError):
         toCamelCase("")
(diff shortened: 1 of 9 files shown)


jjmurre (approved):
LGTM, nice that we now support these 'double-capital' identifiers!

tests/test_naming.py
```diff
@@ -38,6 +39,23 @@ def test_to_snake_case() -> None:
...
+    """Prove that snake and camel case functions are inverses of eachother."""
+    # mind the lower case "i" after "33". It should be made upper case
+    assert to_snake_case(toCamelCase("per_jaar_per_m2")) == "per_jaar_per_m2"
+    assert to_snake_case(toCamelCase("test_name_magic")) == "test_name_magic"
+    assert (
+        to_snake_case(toCamelCase("numbers_33_in_the_middle_44")) == "numbers_33_in_the_middle_44"
+    )
+    assert to_snake_case(toCamelCase("eerste_h_n_id")) == "eerste_h_n_id"
+
+    assert toCamelCase(to_snake_case("numbers33inTheMiddle44")) == "numbers33InTheMiddle44"
```
lbam, line 58: In this case, `toCamelCase` is clearly not the inverse of `to_snake_case`: `"numbers33inTheMiddle44" != "numbers33InTheMiddle44"`.

DonQueso89: good catch!

jjmurre (approved):

PR #479, opened 2023-06-07
Implement logic to recover from failed event messages

JKoetsier:
This new logic makes sure that when a message is redelivered, we can recover from it and the database state after having processed the redelivered message would be the same as if the message would have been processed correctly the first time.


barrydaniels-nl (approved):
Looks good

PR #485, opened 2023-06-14
Bdaniels/view importer

barrydaniels-nl:
* Added features to recognise when a dataset is a view
* Adeded CLI and django management commands to generate the tables for these views.


src/schematools/contrib/django/management/commands/create_views.py
```diff
@@ -0,0 +1,66 @@
...
+def create_views(
+    command: BaseCommand,
+    datasets: Iterable[Dataset],
+    base_app_name: Optional[str] = None,
+) -> None:  # noqa: C901
+    """Create tables for all updated datasets.
+    This is a separate function to allow easy reuse.
+    """
+    errors = 0
+    command.stdout.write("Creating tables")
```
jjmurre: Shouldn't the message be "creating views"?

src/schematools/contrib/django/management/commands/create_views.py
```diff
@@ -0,0 +1,66 @@
...
+    with connection.schema_editor() as schema_editor:
+        for db_table_name, models_group in models_by_table.items():
+            model = max(models_group, key=lambda model: model._dataset.version)
+            if model.is_view():
+                try:
+                    command.stdout.write(f"* Creating view {model._meta.db_table}")
+                    with transaction.atomic():
+                        schema_editor.create_model(model)
+                except (DatabaseError, ValueError) as e:
+                    command.stderr.write(f"  Tables not created: {e}")
```
jjmurre: Same here Table -> View ?

src/schematools/importer/base.py
```diff
@@ -287,6 +293,55 @@ def generate_db_objects(
                 db_schema_name,
             )
 
+    def load_view_sql(self, url: str) -> str:
+        """Load a view SQL file from a URL."""
+        if not url.startswith("http"):
+            raise ValueError(f"View URL {url} is not a valid URL")
+
+        response = requests.get(url)
```
jjmurre, line 304: Normally I would expect here a `raise_for_status`. Or, it that not needed anymore to trigger exception raising?

jjmurre: There are two mechanisms to create a view. One using SQLAlchemy and one using the Django ORM. It seems they do more or less the same? Or, are they complementary?

barrydaniels-nl: Changing the function to use raise_for_status. 

I use the Django ORM in the Django management commands and the SQL Alchemy one in the CLI. It looks to me this is how others have done it.

src/schematools/contrib/django/management/commands/create_views.py
barrydaniels-nl: Fixing

PR #498, opened 2023-09-07
Fix bug for parent tables with two similar relation names

JKoetsier:


src/schematools/events/full.py
```diff
@@ -61,7 +62,8 @@ def parent_id_value(self, prepared_row: dict):
 
     @property
     def parent_fields(self):
-        return [f.name for f in self.parent_table.columns if f.name.startswith(self.relation_name)]
+        pattern = re.compile(f"{self.relation_name}_[a-zA-Z]+")
```
lbam, line 65: None of my business, but...

```suggestion
        pattern = re.compile(f"{re.escape(self.relation_name)}_[a-zA-Z]+")
```

JKoetsier: Didn't know about that method :)

Not needed in this case though, as self.relation_name always consists of letters a-z and possible one or more underscores.

lbam: ... until you hit a bug elsewhere and it doesn't.

jjmurre (approved):
LGTM

PR #499, opened 2023-09-07
Use dataset-specific schema to store temporary full load tables. Update main table relation columns after full load of relation table.

JKoetsier:


jjmurre (approved):
LGTM

PR #501, opened 2023-09-13
Support running DSO-API unittests against schematools

DonQueso89:
This PR adds tools for testing schematools against DSO.

The main idea is that we have a Dockerfile which is used locally (thru docker compose) and in an Github Action, to test schematools against the latest DSO (defined to be the tip of master). Since DSO is only compatible with >= python 3.11, we only run these itegration tests for mentioned python version. The compose file has been updated and modernized a bit to act as an executable for testing.

The Github action DSO-test is triggered for all tag push events, including major version releases. But since we usually release packages "manually", a tag push to master is actually too late. It is at the discretion of the devs to make sure that no minor/patch version that breaks DSO is released.

**WARNING**: DSO is currently not compatible with the newest version of schematools, so merging this will "break" master.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,41 @@
-FROM amsterdam/python:3.9-buster
+# The image is meant to be used as an executable that invokes pytest
+# on schematools by default.
+# To test schematools against the DSO-API, run:
+# docker run <image_name> /app/dso-api/src
+FROM python:3.11-slim-bullseye AS builder
+RUN apt update && apt install --no-install-recommends -y \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,9 +1,41 @@
-FROM amsterdam/python:3.9-buster
```
jjmurre, line 1: Would it be possible to stop using the `amsterdam` base images?

jjmurre (approved):
LGTM

PR #504, opened 2023-09-14
create-views django management command

barrydaniels-nl:


src/schematools/contrib/django/management/commands/create_views.py
```diff
@@ -0,0 +1,195 @@
...
+
+def _is_valid_sql(sql: str) -> bool:
+    """Try and execute the SQL return true if it succeeds, rollback the transaction."""
+    try:
+        with transaction.atomic():
+            sid = transaction.savepoint()
+            with connection.cursor() as cursor:
+                cursor.execute(sql)
+            transaction.savepoint_rollback(sid)
+    except:
```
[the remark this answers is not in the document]
barrydaniels-nl: Ik zal er except Exception: van maken.

src/schematools/contrib/django/management/commands/create_views.py
```diff
@@ -0,0 +1,195 @@
...
+        return False
+    return True
+
+
+def _get_scopes(datasetname: str, tablename: str) -> str:
```
jjmurre, line 38: Looks like this returns a list, so ` -> list[str]`?

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = amsterdam-schema-tools
-version = 5.16.1
+version = 5.16.2
```
jjmurre, line 3: I think this would be a minor bump, because it it not a bug fix.

src/schematools/contrib/django/management/commands/create_views.py
```diff
@@ -0,0 +1,195 @@
...
+    all_scopes = []
+    datasetname = to_snake_case(datasetname)
+    tables = DATASETS.get(name=datasetname).schema.tables
+    for table in tables:
```
jjmurre, line 42: Maybe not for now, but I think this could be implemented somewhat easier with sets.

jjmurre: Something along these lines:
```
def _get_scopes2(datasetname: str, tablename: str) -> frozenset[str]:
    from operator import __or__
    from functools import reduce
    dataset = DATASETS.get(name=datasetname).schema
    table = dataset.get_table_by_id(tablename)
    return table.auth | reduce(__or__, [f.auth for f in table.fields])
```

jjmurre (approved):
LGTM

PR #506, opened 2023-09-18
AB#92275 Circumvent wrong date in bag_panden during export

jjmurre:
The bag_panden table had a historical record with a year of 0001. The dbapi.fetchmany function cannot handle this year, so we need a way to circumvent this problem. Luckily, in psycopg2 we are able to define custom handlers for specific column types. So, we convert datetimes to string, which is done anyway during the export.


barrydaniels-nl (approved):
Ok

PR #534, opened 2023-12-20
AB#101824 Add helper method to userscopes

jjmurre:
To determine if the user an access all fields of a particular table. This new method is needed, because on Azure the postgresql columns can have individual scope-based access. Django collects the full set of columns (select * from) when accessing a FK or M2M relation. This leads to permission error at the database level when accessing the API.

So, more specific access of columns is needed,
hence the extra helper method on userscopes.

### DIFF
--- a/tests/test_permissions_auth.py
+++ b/tests/test_permissions_auth.py
@@ -154,3 +154,13 @@ class TestFieldAccess:
         assert user_scopes.has_field_access(table.get_field_by_id("volgnummer")) == expect
         assert user_scopes.has_field_access(table.get_field_by_id("identificatie")) == expect
         assert user_scopes.has_field_access(table.get_field_by_id("registratiedatum")) == expect
+
+    def test_has_table_fields_access(self, id_auth_schema):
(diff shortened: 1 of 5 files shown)


tests/test_permissions_auth.py
```diff
@@ -154,3 +154,13 @@ def test_profile_field_inheritance_from_dataset(
         assert user_scopes.has_field_access(table.get_field_by_id("volgnummer")) == expect
         assert user_scopes.has_field_access(table.get_field_by_id("identificatie")) == expect
         assert user_scopes.has_field_access(table.get_field_by_id("registratiedatum")) == expect
+
+    def test_table_fields_access(self, id_auth_schema):
```
roelof-adam, line 158: wellicht toevoegen 'has' om 1 op 1 te matchen met functienaam. test_has_table_fields_access

roelof-adam (approved):
LGTM

tests/test_permissions_auth.py
jjmurre: Good point, changed!

PR #540, opened 2024-01-22
Fix storing the display_field on the datasettable

jjmurre:
A stringified tuple was saved to the database.
Seems to be some kind of copy/paste error.


barrydaniels-nl (approved):
Ok

PR #544, opened 2024-02-06
Do not write export files if not columns are selected

jjmurre:
Sometimes zero columns are selected. Only open data is exported, so when columns are protected by a scope, those columns are not exported.

Furthermore, the temporal aspect was not added to the jsonlines export, so the whole history was exported.


barrydaniels-nl (approved):
Ok

PR #545, opened 2024-02-07
Fix premature leave of the export loop.

jjmurre:
The loop through the tables of a dataset during export was prematurely being left when a table was empty (because it has no open data).


barrydaniels-nl (approved):
Ok

PR #547, opened 2024-02-07
Added temporal indexes

barrydaniels-nl:


jjmurre (approved):
LGTM

src/schematools/factories.py
```diff
@@ -407,6 +408,36 @@ def _build_geo_indexes(
     ]
 
 
+def _build_temporal_indexes(
+    table_object: Table, dataset_table: DatasetTableSchema, db_table_name: str
+) -> list[Index]:
+    """Creates an index on temporal fields."""
+    if dataset_table._temporal_range_field_ids:
```
jjmurre, line 415: One question. Why the same if statement twice?

src/schematools/factories.py
```diff
@@ -407,6 +408,36 @@ def _build_geo_indexes(
     ]
 
 
+def _build_temporal_indexes(
```
jjmurre, line 411: If possible, remove the leading `_`, because now we are using an internal attribute of the class which is considered dangerous.

barrydaniels-nl: Done

PR #548, opened 2024-02-29
AB#108238 Add `id_field` to Datasettable

jjmurre:
Is needed for geosearch. In geosearch the `id_field` is now hardcoded on `id`. But, for singular PK table, a different id for the PK should be possible.


barrydaniels-nl (approved):
Looks good

PR #549, opened 2024-03-04
Fix casing of the id_field.

jjmurre:
Because it is a databasefield and it is directly used in sql queries, it is better to snake_case it.


lennardstam (approved):
OK

PR #551, opened 2024-04-30
Yrob/add AB#114790 units to fields

Yrob:
[AB#114790](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/114790) Add unit property from amsterdam-schema to field type.


jjmurre (approved):
LGTM

PR #554, opened 2024-07-08
Remove BenK EventsProcessor code, this is no longer used.

vdboor:
Confirmed by @peteradrichem

This code was no longer working since the benk-dataset was removed in: https://github.com/Amsterdam/amsterdam-schema/pull/1032 This also removes the `schema import events` CLI command.


barrydaniels-nl (approved):
LGTM

PR #557, opened 2024-07-11
Remove deprecated APIs for the next major release

vdboor:
* schematools.utils was replaced by the loader API.
* is_relation_temporal was only used internally.


barrydaniels-nl (approved):
Ok

PR #560, opened 2024-08-01
Improved testing for mandatoryFilterSet

vdboor:
* Test what happens when a scope is missing
* Fix BRP_R profile -> BRP_RNAME so these auth/profile roles don't overlap.
* Added brp_schema fixture, change how brp_dataset fixture is constructed.


barrydaniels-nl (approved):
Ok

PR #563, opened 2024-09-26
Fix `schema permission apply` and add sequence support.

vdboor:
* Adds grants to sequences
* Fixes nested/through tables.

The old `get_all_dataset_scopes()` walked twice over nested/through tables, and gave it the dataset permissions too (instead of the field permissions). This was visible when refactoring the code, and unit tests confirmed. Hence the list of generated GRANT statements is now included in the tests.


MaguellSandifort (approved):

vdboor:
Gecheckt, in de praktijk heeft dit alleen effect op `brk.kadastraleobjecten.soortCultuurBebouwd`. En brk2 is de vervanger hiervan.

PR #567, opened 2025-01-09
Add compatibilty for python 3.13

jasper-jager:
Updated all references to python 3.13, also added python 3.12 to the test matrix.

Since the last successful test workflow on Github the `ubuntu:latest` image was updated from 22.04 to 24.04, which required an update to the gdal package. Should we pin the version of the test container to prevent this from happening again?


vdboor (approved):
LGTM

PR #568, opened 2025-01-16
Inline strtobool to replace deprecated distutils

jasper-jager:
As per python 3.12 distutils is deprecated. The only use of this was strtobool which has been reimplemented inline.


vdboor (approved):
nice!

PR #569, opened 2025-01-20
Update CLI to accept environment variables and correct Amsterdam Schema

jasper-jager:
The DSO API is using a mangled and outdated version of the Amsterdam Schema to generate mock data on initializing the database in docker compose. To fix this, schema tools has been updated to accept datasets from Amsterdam Schema to generate mock data. 

It's possible to generate mock data using positional arguments or by setting the environment variable `DATASETS_LIST`, which is also used in the DSO API.


vdboor (approved):
LGTM!

PR #587, opened 2025-03-31
Remove version number from model name

jasper-jager:
Removed major version number from version name as it causes issues in the DSO-API for now. We will need this once we'll start to support multiple versions of a model, but for now this causes to much required changes in DSO-API.

### DIFF
--- a/src/schematools/contrib/django/factories.py
+++ b/src/schematools/contrib/django/factories.py
@@ -144,7 +144,7 @@ def get_model_name(table_schema: DatasetTableSchema) -> str:
     # Using table_schema.python_name gives UpperCamelCased names, the old format is kept here.
     # This also keeps model names more readable and recognizable/linkable with db table names.
     model_name = to_snake_case(table_schema.id)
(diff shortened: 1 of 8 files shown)


src/schematools/contrib/django/factories.py
```diff
@@ -144,7 +144,7 @@ def get_model_name(table_schema: DatasetTableSchema) -> str:
     # Using table_schema.python_name gives UpperCamelCased names, the old format is kept here.
     # This also keeps model names more readable and recognizable/linkable with db table names.
     model_name = to_snake_case(table_schema.id)
-    return f"{model_name}_{table_schema.version.vmajor}"
+    return f"{model_name}"
```
pstokkink, line 147: Maybe update the docstring too...

pstokkink (approved):
with a comment

src/schematools/contrib/django/factories.py
vdboor: or: 
```suggestion
    return model_name
```

PR #600, opened 2025-04-24
fixed bug in geopackage exporter to handle both SQLAlchemy engine and…

barrydaniels-nl:
fixed bug in geopackage exporter to handle both SQLAlchemy engine and connection objects


pstokkink (approved):
LGTM

PR #617, opened 2025-05-22
Make import schemas atomic and fix empty migrations

jasper-jager:
Import schemas was not atomic, which caused issues when a migration failes. The schema data was already updated, which results in a mismatch between the schema in datasets_dataset and the actual tables.

For metadata changes an empty migration could be generated (with only sql comments). If this happens we'll skip the migration.


pstokkink (approved):
LGTM

PR #621, opened 2025-06-10
Add ROLE.filtered roles for every created permission

vdboor:
This allows Terraform to link DSO-API to "role.filtered", while the SQL user gets "role" without the .filtered.
This way, DSO-API can access the mandatoryFilterSets columns, but the SQL user can't.

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -114,9 +114,13 @@ class TestReadPermissions:
         grants = _filter_grant_statements(caplog)
         assert grants == [
             'CREATE ROLE "scope_brk_encoded"',
+            'CREATE ROLE "scope_brk_encoded.filtered" IN ROLE "scope_brk_encoded"',
             'CREATE ROLE "scope_brk_rid"',
-            'CREATE ROLE "scope_brp_rname.filtered"',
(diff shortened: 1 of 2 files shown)


tests/test_permissions.py
```diff
@@ -113,6 +113,14 @@ def test_all_profile_permissions(
...
         grants = _filter_grant_statements(caplog)
         assert grants == [
+            'CREATE ROLE "scope_brk_encoded"',
+            'CREATE ROLE "scope_brk_encoded.filtered" IN ROLE "scope_brk_encoded"',
+            'CREATE ROLE "scope_brk_rid"',
+            'CREATE ROLE "scope_brk_rid.filtered" IN ROLE "scope_brk_rid"',
+            'CREATE ROLE "scope_brp_rname"',
+            'CREATE ROLE "scope_brp_rname.filtered" IN ROLE "scope_brp_rname"',
+            'CREATE ROLE "scope_fp_md"',
+            'CREATE ROLE "scope_fp_md.filtered" IN ROLE "scope_fp_md"',
```
pstokkink, line 123: Why are the un.filtered roles new in this list?

pstokkink (approved):
👍

PR #643, opened 2025-09-10
Add migrate tables flag to import schemas

jasper-jager:
Add flag to be able to import schemas without trying to create migrations for non existing tables. This is used in Amsterdam Schema API to have access to all schemas in a local database


JonaBenja (approved):
LGTM!

PR #648, opened 2025-09-25
Jona/filter datasetschema on scope

JonaBenja:
- Added forgotten check for Dataset access
- Only add scope "OPENBAAR" when it is not already provided
   - Because we are comparing Scopes in a set, this would not cause issues, but we just don't need two OPENBAAR scopes

- Created a `.post1 ` version, following this documentation: https://packaging.python.org/en/latest/discussions/versioning/


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# 2025-09-24 (8.3.6.post1)
```
JonaBenja, line 1: Ho wacht de datum klopt niet hier

pstokkink (approved):

PR #649, opened 2025-09-25
Delete datasets when dataset.json schema is deleted

lotte-amsterdam:
- Compare datasets and delete datasets if missing
- Tests for comparing and deleting datasets
- Added management commands to import scopes and publishers

### DIFF
--- a/tests/django/fixtures.py
+++ b/tests/django/fixtures.py
@@ -147,7 +147,7 @@ def dataset_library(
     gebieden_schema: DatasetSchema,
     parkeervakken_schema: DatasetSchema,
     afval_schema: DatasetSchema,
-) -> Dataset:
+) -> dict[str, Dataset]:
     """Create a dataset library"""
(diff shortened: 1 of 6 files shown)


tests/django/fixtures.py
```diff
@@ -140,3 +140,20 @@ def woningbouwplannen_dataset(woningbouwplannen_schema: DatasetSchema) -> Datase
...
     """Create Metaschema v3 dataset."""
     return Dataset.create_for_schema(metaschemav3_schema)
+
+
+@pytest.fixture
+def dataset_library(
+    gebieden_schema: DatasetSchema,
+    parkeervakken_schema: DatasetSchema,
+    afval_schema: DatasetSchema,
+) -> Dataset:
```
pstokkink, line 150: Dit return type klopt niet, het is een `dict[str, Dataset]`

pstokkink (changes requested):
Paar kleine dingetjes:

pstokkink (approved):
LGTM

PR #726, opened 2026-06-15
Upgrade toolchain

pstokkink:
* Upgrade toolchain to use `uv`.
  * Pin all dependencies
  * Clean up docker setup
  * Update Makefile
  * Update github workflows (deplic -> licensecheck, so we can use uv everywhere)
  * Update README.md
  * Dependabot toegevoegd


jasper-jager (approved):
Nice!

PR #747, opened 2026-07-28
[BUGFIX] escape percentage char

pstokkink:
Generated SQL included the following:
```
-- Create constraint benkagg_v1_handelsregisterkvk_v4_identificatie_not_contains_slash on model table
--
ALTER TABLE "table_v3" ADD CONSTRAINT "table_v4_identificatie_not_contains_slash" CHECK (NOT ("identificatie"::text LIKE '%/%'));
--
```
This errored in psycopg's cursor.execute, because it tries to interpolate `%/`, which is invalid.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# 2026-07-28 (9.11.1)
+
+* Bugfix: escape %-sign in generated SQL.
+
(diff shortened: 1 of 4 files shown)


JonaBenja (approved):
OK

PR #749, opened 2026-07-29
Bugfixes

pstokkink:
* Allow the CLI to load without the optional Databricks dependencies; only the `ingest` command now requires that extra when used.
* Report validator exceptions as validation errors instead of aborting validation.
* Improve test database and SQLite cleanup, and add coverage for import, export, CLI, and permission flows.


lotte-amsterdam (commented):
LGTM

lotte-amsterdam (approved):

